PR #25, opened 2026-04-12
fix: prevent AutoSync race condition by setting running flag early

gnarg:
Qwen thinks there's a race condition here, and this is the fix.


benders (approved):
[@claude]: > Correct fix. Moving `running = true` before the first `await` closes the race window. The `try/finally` wrapper correctly resets the flag on all exit paths. CI tweak is unrelated but harmless.

PR #26, opened 2026-04-12
fix(auth): validate hex format for enc: password decoding

gnarg:
Qwen thinks encoded passwords should be validated for correct hex.


benders:
This one doesn't make any sense to me. If the password is invalid, it won't be in the database, so it'll be failed at with "Wrong username or password" anyway.

Issue #2, opened 2026-04-11
Stream player: show source instance, format, and bitrate


benders:
Player UI should display which instance the track streams from, audio format (mp3/flac/etc), and bitrate. Info available from track_sources.

_Migrated from yak poutine-15c3_

PR #33, opened 2026-04-12
#2 track details in player view

gnarg:
Something simple to try out some agent workflows.

### DIFF
--- a/frontend/src/components/player/PlayerBar.tsx
+++ b/frontend/src/components/player/PlayerBar.tsx
@@ -110,7 +110,7 @@ export function PlayerBar() {
   }
 
   return (
-    <div className="fixed bottom-0 left-0 right-0 h-20 bg-player border-t border-border flex items-center px-4 gap-4 z-50">
+    <div className="fixed bottom-0 left-0 right-0 h-20 bg-player border-t border-border flex items-center px-8 gap-4 z-50">
       <audio
         ref={audioRef}
         preload="auto"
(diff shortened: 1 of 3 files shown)


benders (commented):
@claude:

> **`hub/src/routes/subsonic.ts` `buildSong()`:** 🔴 bug: \`suffix\` never sent. \`TrackRow.format\` exists but \`buildSong\` doesn't map it. Fix: add \`suffix: row.format ?? undefined\`. Frontend will always show nothing for format.
> 
> **`hub/src/routes/subsonic.ts` L464, L526, L601:** 🔵 nit: Same 4-line \`instance_name\` subquery copy-pasted 3x (same pattern as existing \`bitrate\`/\`size\`). Now 4 correlated subqueries per track row — fine for now but worth a CTE or lateral join if query plans get heavy.
> 
> **`frontend/src/components/player/PlayerBar.tsx`:** 🟡 risk: Removed \`w-56\` fixed width from track info container — track info now grows unbounded with long titles and could squeeze controls on narrow viewports. Consider adding \`max-w-xs\` or similar cap.
> 
> **`frontend/src/components/player/PlayerBar.tsx` seek bar div:** 🔵 nit: Added \`max-w-xl\` but parent already has \`max-w-xl\`. Redundant, harmless.
> 
> **`frontend/src/lib/subsonic.ts`:** Clean. Interface + parser additions look good.
> 
> ---
> Main blocker: fix the missing \`suffix\` mapping in \`buildSong\` before merging.

gnarg:
> hub/src/routes/subsonic.ts buildSong(): 🔴 bug: `suffix` never sent. `TrackRow.format` exists but `buildSong` doesn't map it. Fix: add `suffix: row.format ?? undefined`. Frontend will always show nothing for format.

Neither Qwen nor I have any idea what this is about. "Frontend will always show nothing for format." is provably false, since it works fine for me locally. The rest of the issues have been addressed.

benders (approved):

PR #76, opened 2026-04-19
feat: minimum api-version enforcement

loving-grace:
Implements #5


gnarg:
Too disorganized, mixed with activity stream work, removed useful log lines.

PR #110, opened 2026-04-27
Reversible passwords + Subsonic u+t+s (closes #106)

### LINKED ISSUE #106: Should we support token + md5 Subsonic authentication?
Modern Subsonic clients rely on a shared secret for the password. This is cool in that it works securely over plain HTTP, but sucks because it means storing the password in plaintext on the server. 

To avoid having the passwords just sitting there for random passers by, Navidrome encrypts them using `PasswordEncryptionKey`. Right now we don't specify this (env var would be `ND_PASSWORDENCRYPTIONKEY`), so it uses a global default. 

Instance security isn't a big deal, since everyone hosts their own instance. But we should probably adopt a similar process for Poutine if we want Subsonic auth to work as expected. That means a breaking change where we re-initialize the password table to be reversable instead of hashed, and implement some kind of encryption key. For the key injection, creating it on first-boot and placing it in the filesystem probably makes the most sense.

Migration can include manual steps for admins to take. We are early in deployment now.

benders:
## Summary

- Replace Argon2id user-password hashing with AES-256-GCM reversible storage. Encryption key auto-generated on first boot at `$POUTINE_PASSWORD_KEY_PATH` (default `./data/poutine_password_key`, mode 0600).
- Add `u+t+s` (MD5 token+salt) auth to `/rest/*` and `/proxy/*` alongside the existing `u+p` path.
- Migrate the SPA to Subsonic `u+t+s`: `/admin/login` returns `subsonicCredentials`, `subsonicFetch`/`streamUrl`/`artUrl` compute a fresh salt+md5 per request, and the JWT path on `/rest/*` is gone (as is the SPA's mid-track 401 retry, which existed only to bridge JWT expiry).
- Drop the `argon2` runtime dependency. Add `js-md5` to the frontend.

Five focused commits map to the five steps of the implementation; each was committed only after `pnpm typecheck` + `pnpm test` were green.

## Breaking change — passwords reset on upgrade

The migration converts `users.password_hash` → `users.password_enc` and sets `password_enc=''` for every row (Argon2 hashes are not reversible). Operators must:

1. Set `POUTINE_OWNER_USERNAME` + `POUTINE_OWNER_PASSWORD` so `seedOwner()` recovers the owner row on boot.
2. Re-set passwords for any other users via the admin UI after logging in.
3. Back up `data/poutine_password_key` alongside the SQLite DB. **Lose the key file = every password is unrecoverable.**

`README.md` has the operator-facing upgrade notes; `docs/authentication.md` has the developer-facing detail.

## Test plan

- [x] `pnpm typecheck` — clean.
- [x] `pnpm --filter @poutine/hub test` — 278 tests pass (was 270; +8 new for crypto, migration, seed-owner recovery, u+t+s on `/rest/*` and `/proxy/*`, login response shape).
- [x] `pnpm --filter @poutine/frontend test` — 19 tests pass.
- [x] `pnpm --filter @poutine/frontend build` — clean.
- [x] **Manual browser testing recommended before merge:** log in via the SPA, browse / play / seek a track, verify cover art loads, confirm logout clears `subsonicUser`/`subsonicPass` from localStorage. Try a third-party Subsonic client (DSub / Symfonium) with `u+t+s` auth.
- [ ] Verify upgrade path on a snapshot of a real DB: copy `data/poutine.db` to a scratch dir, boot 0.4.0 with the same env, confirm owner login works, confirm guest user passwords are blank and reset via UI works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ config/peers.yaml
 *.db-wal
 *.db-shm
 *.pem
+poutine_password_key
 *.data.json
 artists.json
 coverage/
--- a/README.md
+++ b/README.md
@@ -66,6 +66,14 @@ docker compose up -d hub
 
 Running containers use the compiled image, not live source — rebuild is required after any code change.
 
+#### Upgrading to 0.4.0 — password reset required
+
+0.4.0 changes how user passwords are stored (Argon2id → AES-256-GCM, reversible storage so Subsonic `u+t+s` works). All existing passwords are wiped on the upgrade — every user must have their password re-set.
+
+1. Make sure `POUTINE_OWNER_USERNAME` and `POUTINE_OWNER_PASSWORD` are set in your env. On boot, the hub recovers the owner row by re-encrypting `POUTINE_OWNER_PASSWORD`.
+2. After the hub starts, log in as the owner and re-create or re-set passwords for any other users via the admin UI.
+3. Back up `data/poutine_password_key` (auto-generated on first boot, mode 0600) alongside your SQLite DB. **Losing the key file makes every stored password unrecoverable.** Override the path with `POUTINE_PASSWORD_KEY_PATH` if needed.
+
 ### Resetting the owner password
 
 Owner seeding only runs on first boot (when `users` is empty). To reset a password while the hub is running:
@@ -99,7 +107,7 @@ Navidrome's admin-bootstrap env vars only run on a fresh volume. To force a rese
 
 | Layer     | Tech                                                      |
 |-----------|-----------------------------------------------------------|
-| Hub       | TypeScript, Fastify, better-sqlite3, jose (JWT), argon2   |
+| Hub       | TypeScript, Fastify, better-sqlite3, jose (JWT), AES-GCM  |
 | Frontend  | React 19, Vite, Tailwind CSS, Zustand, TanStack Query     |
 | Per-peer  | Navidrome (Subsonic / OpenSubsonic API)                   |
 | Transcode | FFmpeg (via Navidrome, never on the hub)                  |
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -7,42 +7,36 @@ Poutine has three authentication mechanisms, each scoped to a different API surf
 | Surface           | Prefix          | Mechanism                                                                             |
 |-------------------|-----------------|---------------------------------------------------------------------------------------|
 | Admin             | `/admin/*`      | JWT (cookie + `Authorization: Bearer` header)                                         |
-| Subsonic (JSON)   | `/rest/*`       | JWT **or** legacy Subsonic `u`+`p` query params                                       |
+| Subsonic (JSON)   | `/rest/*`       | Subsonic `u+p` (plaintext / `enc:<hex>`) **or** `u+t+s` (MD5 token+salt)              |
 | Subsonic (binary) | `/rest/stream`, `/rest/getCoverArt` | Same as Subsonic JSON, but errors use HTTP status codes, not Subsonic envelopes |
-| Proxy             | `/proxy/*`      | Unified: Ed25519 (peers) → JWT (SPA) → Subsonic `u`+`p` (3rd-party), tried in order |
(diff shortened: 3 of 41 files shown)


benders:
@claude: code review

> **Overview:** crypto primitives correct, migration contained, docs coherent. Main blocker-adjacent concern is plaintext password lifetime in localStorage. Items below ranked by severity.

> **1. Plaintext password lifetime in browser localStorage (security regression).** `/admin/login` returns `subsonicCredentials.password` in plaintext; SPA persists to `localStorage` indefinitely. Pre-#106, an XSS leaked a 15-min access token; now it leaks the user's password until manual rotation. PR description claims "same threat surface as the JWT" — not quite true (token has TTL, password doesn't). Consider `sessionStorage` (cleared on tab close), or at minimum add an explicit "Security trade-offs" note to `docs/authentication.md` so future contributors don't lose track.

> **2. Username enumeration via timing.** `hub/src/auth/subsonic-auth.ts::lookupAndVerify` short-circuits with `return null` when the user row is missing — no dummy decrypt+md5 performed. AES-decrypt path on `u+p` is measurable. Minor; future hardening.

> **3. No salt-length validation.** `readSubsonicCreds` and `hub/src/proxy/auth.ts` accept any non-empty `s`. Subsonic spec requires salt ≥ 6 chars. Reject short salts as defense-in-depth/spec compliance.

> **4. `streamUrl()` / `artUrl()` fallback when unauthenticated.** Both fall back to `new URLSearchParams({ v, c })` when `getSubsonicCreds()` is null. Resulting URL 401s anyway; cleaner to return `null` and let callers handle it (PlayerBar already handles `currentStreamUrl === null`).

> **5. Dead refs in PlayerBar.** `baseOffsetRef` and `pendingBaseOffsetRef` are still declared and reset after the 401-retry deletion. Verify they're still used by the transcoded-offset seeker — drop if not.

> **6. `seed-instances.ts` placeholder interaction.** `__placeholder__` → `''` combined with the new "treat `__system__` as no real users" rule in `seedOwner` is subtle. Add a comment in `seedOwner` pointing at `seedSyntheticInstances` so the interaction is discoverable.

> **7. Manual test items unchecked.** PR description's two manual items (browser play/seek, real-DB upgrade dry-run) are still unchecked. Given the breaking migration, run the snapshot upgrade before merge.

> **Strengths:** AES-256-GCM with 12-byte IV / 16-byte tag / `timingSafeEqual`, tamper + wrong-key tests; key file 0600, malformed rejection, gitignored, prominent "lose the key = lose all passwords" callout; migration test exercises a hand-built pre-#106 schema; `useMemo([currentTrack?.id])` on `streamUrl()` plus a "Maximum update depth" regression test; docs across `authentication.md`, `hub-internals.md`, `system-architecture.md`, `opensubsonic.md`, README all updated coherently with no stale Argon2 references.

> **Verdict:** merge after addressing #1 (at minimum the docs note) and running the manual upgrade dry-run. #2–#6 are follow-ups.

benders:
@claude: testing-finding follow-up

> **Finding:** the `/rest/*` redirect-to-login branch in `frontend/src/lib/subsonic.ts::subsonicFetch` only handles codes **40** and **41**. OpenSubsonic defines five credential-related codes that the SPA should treat identically: **40, 41, 42, 43, 44**, plus **10** when caused by missing creds. Code **50** is authorization (insufficient privilege), not authentication — must NOT redirect.

> Full audit:
> - 10 — Required parameter missing → redirect (implies creds absent)
> - 20 / 30 — version mismatch → no redirect
> - 40 — Wrong username or password → redirect ✅ already handled
> - 41 — Token auth not supported (LDAP) → redirect ✅ already handled
> - 42 — Auth mechanism not supported → **redirect (missing)**
> - 43 — Multiple conflicting auth mechanisms → **redirect (missing)**
> - 44 — Invalid API key → **redirect (missing)**
> - 50 — Not authorized for operation → no redirect (authz)
> - 60 / 70 — trial / not found → no redirect

> Filed as #111 with acceptance criteria + unit-test ask. Suggest fixing in this PR while we're in the auth-redirect codepath, or as a small follow-up.

PR #119, opened 2026-04-28
fix: prevent unified_*.id PK collisions on metadata variants

### LINKED ISSUE #118: fix: unified entity PK collisions across upstream metadata variants
Live testing in \`test/local-cluster\` produced two distinct \`UNIQUE constraint failed\` errors on \`mergeLibraries\`. Both stem from \`generateReleaseId\`/\`generateTrackId\` not including enough of the natural key in their hash, so legitimate-but-divergent upstream rows collide on the unified PK.

Supersedes #117.

## 1. unified_releases.id collision (non-MBID, same RG, different track counts)

Repro: Leonard Cohen / *Old Ideas*. Two of the ten m4a files have inconsistent \`date\` tags (tracks 03 \`2011-12-27\`, 04 \`2012-01-16\`, the rest \`2012-01-27\`). Navidrome groups by date and splits the single physical album into 3 instance_albums named \"Old Ideas\" with track counts 8/1/1. They land in the same release group. \`generateReleaseId(name, rgId, null)\` ignores \`track_count\` → all three hash to the same id.

**Observed log**: \`existingRow.track_count=1\`, new insert \`track_count=8\`, identical id.

**Fix**: include \`trackCount\` in the non-MBID branch of \`generateReleaseId\`.

## 2. unified_tracks.id collision (same recording MBID across releases)

Repro: Chemical Brothers / \"Setting Sun\". Single (322s) and album version on *Dig Your Own Hole* (329s) share recording MBID \`7a7d7fb7-075b-4fdc-8c5e-9b5e03ee00b3\` — this is normal MusicBrainz: a *recording* MBID is unique per recording but can appear on many releases. \`generateTrackId\` MBID branch hashes only \`(\"track\", mbid)\` → second insert collides.

**Fix**: include \`releaseId\` in the MBID branch of \`generateTrackId\`. (Deterministic releaseId keeps federation peers in sync.)

## Diagnostics

Every \`unified_*\` insert in \`merge.ts\` is now wrapped in a try/catch that logs the offending row, the prior \`existingRow\` that holds the colliding id, and the source instance/album/track ids before rethrowing. These stay in place; future merge issues should be self-diagnosing from the logs.

## Tests

- \`hub/test/id-generator.test.ts\`: trackCount in non-MBID release id; recording MBID across distinct releases.
- \`hub/test/merge.test.ts\`: end-to-end regression for both scenarios.

benders:
This is a follow-up fix for issues in #103 . Closes #118.

## Summary
- \`generateReleaseId\` non-MBID branch now folds \`trackCount\` into the hash, so one upstream album split by Navidrome into multiple instance_albums (Leonard Cohen / *Old Ideas*) no longer collides on \`unified_releases.id\`.
- \`generateTrackId\` MBID branch now scopes by \`releaseId\`, so a single recording MBID that legitimately appears on multiple releases (Chemical Brothers / "Setting Sun" single + album) no longer collides on \`unified_tracks.id\`. Deterministic releaseId keeps federation peers in sync.
- Every \`unified_*\` insert in \`merge.ts\` is wrapped with structured logging (offending row, prior \`existingRow\`, source ids) so future PK collisions are self-diagnosing.

## Test plan
- [x] \`pnpm typecheck\`
- [x] \`pnpm test\` — 315 tests pass, including new id-generator and full-merge regressions for both scenarios
- [x] Re-run \`test/local-cluster\` against the cd-rips library that previously failed; confirm sync completes
- [x] Verify federation peers still resolve consistent ids after the hash change (IDs for non-MBID releases and all MBID-tagged tracks shift on next merge — same precedent as #103)


loving-grace:
@claude (review): checked out `fix/unified-id-collisions-118`, ran `pnpm typecheck` and `pnpm test` — both clean (363 tests). Code matches the description; a few concerns:

> 1. **Undocumented third hash change.** The PR description covers `generateReleaseId` (trackCount) and `generateTrackId` (releaseId scoping), but commit 3f28176 also changes `generateTrackSourceId` to include `instance_track_id`. That is a separate schema-affecting hash change (justified by the alt-take scenario in the new test), and it should be called out in the PR body alongside the existing federation-id-shift caveat. Same precedent as #103 applies to it too.
>
> 2. **Federation sync claim is too strong for the trackCount change.** The PR says "Deterministic releaseId keeps federation peers in sync." For MBID-tagged releases this is still true. For non-MBID releases, folding `trackCount` into the hash means peers with different track counts for the same physical album (one missing file, a different rip, an in-progress import) now compute *different* `unified_releases.id`s and will not unify across peers. That is arguably the right correctness behavior, but it is a federation-visible regression vs. the previous `(rgId, nameNormalized)`-only hash, and worth being explicit about.
>
> 3. **UX side-effect of the Old Ideas fix.** The new merge test asserts that the trackCount=1 splits and the trackCount=8 split produce two distinct `unified_releases` both named "Old Ideas". That is correct as a crash fix, but on a single instance the user will now see one physical album as two "Old Ideas" rows in the UI until the upstream metadata is cleaned up. Probably acceptable; might warrant a TODO or a follow-up issue to collapse same-RG / same-name / non-MBID variants in the presentation layer.
>
> 4. **Minor test-style nit.** In `merge.test.ts`, the new "same recording MBID on two different releases" test uses `expect(new Set(tracks.map(t => t.id))).toHaveProperty("size", 2)` while the surrounding tests use `expect(new Set(...).size).toBe(2)`. Harmless, just inconsistent.
>
> Logic of the hash changes themselves looks right: the MBID track branch correctly scopes by `releaseId` only (title/track-number variations across releases collapse to one row per release, as the new "stable for same MBID on same release" test verifies), and the non-MBID release branch correctly leaves the MBID branch alone so cross-branch collisions remain vanishingly unlikely. Diagnostic logging in `merge.ts` is thorough and should make the next PK collision self-diagnosing as advertised.

PR #124, opened 2026-04-29
Model peers as Subsonic MusicFolders (#123)

### LINKED ISSUE #123: Model different peers (including local) as "MusicFolders" in the Subsonic API
As part of the goal to adhere to the Subsonic API, each of the active peers including local, can be modeled as a "MusicFolder". This allows other API calls to optionally restrict results to only items on that peer, and moves the concept of albums by peer from being something in the web UI to being something that will work with 3rd party clients as well.

Building the list under "Albums" in the web UI should use the `getMusicFolders` call to get the list, with names and numeric IDs. Then clicking into one of the subsets (like "Local") should use the optional `musicFolderId` parameter to `getAlbumList2` in order to get the correct subset.

Test plan should include checking with at least one 3rd party client to see how the folders are presented.

benders:
## Summary
- Each instance (local + active peers) is now exposed as a Subsonic MusicFolder via `getMusicFolders`, keyed on a stable monotonic int (`instances.musicfolder_id`).
- `getAlbumList2` honors the standard `musicFolderId` param; legacy `instanceId` preserved as an alias.
- Sidebar + AlbumsPage migrated onto `getMusicFolders` / `musicFolderId`; routes are now `/library/folder-<id>` and the hardcoded "Local" entry is gone (Local appears as a folder).

Closes #123

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (hub + frontend, 311 + 45 passing)
- [x] `pnpm test:federation` — 75/75 passing across hub-a/b/c, incl. new assertions: hub-a sees ≥3 folders with unique int ids; `musicFolderId` filter isolates the local folder
- [x] `test/subsonic-compat/test_compat.py` — new cases (`test_get_music_folders`, `test_get_music_folders_includes_self_and_peers`, `test_get_album_list2_music_folder_filter`) pass on every hub
- [x] Manual 3rd-party client (DSub or Submariner): folder picker shows peer names; selecting one restricts album browse

### DIFF
--- a/docs/hub-internals.md
+++ b/docs/hub-internals.md
@@ -105,6 +105,7 @@ Contract: [federation-api.md](federation-api.md). Read before modifying `/federa
 - **Catalogs may diverge between hubs** — each hub builds its own catalog by reading Navidrome instances directly. No fan-out re-export risk.
 - **`GET /admin/peers`** does live health checks: parallel `fetch` to each peer's `/api/health` with a 5 s `AbortController` timeout. `status` from live reachability; `lastSeen` from DB.
 - **`AutoSyncService`** (`hub/src/services/auto-sync.ts`) polls Navidrome every 30 s. When Navidrome's `lastScan` > `instances.last_synced_at` for `'local'`, runs `syncLocal` + `mergeLibraries`. Skips when already running (boolean lock) or when Navidrome is scanning. Wired via `onReady`/`onClose` in `server.ts`.
+- **`instances.musicfolder_id`** (issue #123) is a stable monotonically-assigned int that surfaces each instance as a Subsonic MusicFolder. Subsonic clients require integer folder ids but `instances.id` is a UUID (or `'local'`); this column is the int↔UUID bridge. Assigned in `seedSyntheticInstances` via `MAX(musicfolder_id)+1`; backfilled on upgrade in `db/client.ts` ordered by `created_at`. Resolve via `SELECT id FROM instances WHERE musicfolder_id = ?` — never reuse a value, never recompute on the fly.
 
 ### Versioning
 
--- a/docs/opensubsonic.md
+++ b/docs/opensubsonic.md
@@ -60,7 +60,7 @@ All endpoints support both GET and POST, with and without the `.view` suffix (e.
 
 | Endpoint            | Status          | Notes                                                              |
 |---------------------|-----------------|--------------------------------------------------------------------|
-| `getMusicFolders`   | Implemented     | Returns a single static folder (`id: 1, name: "Music"`)            |
+| `getMusicFolders`   | Implemented     | One folder per known instance (local + active peers); `id` is the stable `instances.musicfolder_id` (issue #123) |
 | `getIndexes`        | Implemented     | Returns artist index from unified library; ignores `musicFolderId` |
 | `getMusicDirectory` | NOT IMPLEMENTED |                                                                    |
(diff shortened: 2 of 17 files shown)


benders:
@claude code review:

> ## Overview
> Replaces the static `getMusicFolders` stub with one folder per known instance (local + peers), keyed on a stable monotonic int (`instances.musicfolder_id`). `getAlbumList2` honors the standard `musicFolderId` (legacy `instanceId` kept as alias). Sidebar + AlbumsPage migrated off `/admin/peers` summary onto `getMusicFolders`. Renames "Local Navidrome" → "Local" in user-facing strings.
>
> ## Strengths
> - Stable, monotonic, non-reused folder ids — correct semantics for a Subsonic catalog identifier.
> - Backfill ordered by `created_at` → deterministic on upgrade.
> - One shared React Query (`["musicFolders"]`) feeds both Sidebar and AlbumsPage with a 60 s stale window; no double-fetch.
> - Test coverage solid: hub unit, federation script (≥3 folders, unique int ids, scoped filter), Python compat suite (self+peers and subset-of-unfiltered).
> - Unknown `musicFolderId` returns empty `albumList2` rather than an error — matches Subsonic client expectations.
> - Docs updated (`docs/hub-internals.md`, `docs/opensubsonic.md`).
>
> ## Issues
>
> ### Medium
> - **Peer folder names are UUIDs by default.** `seedSyntheticInstances` inserts peer rows with `name = peer.id` (`hub/src/library/seed-instances.ts:58`). A new operator who registers a peer without setting a friendly name will see a UUID in the Subsonic folder picker. Suggest `peer.name ?? peer.id` at insert and a refresh on sync.
> - **Schema vs. migration drift.** `hub/src/db/schema.sql:38` declares `musicfolder_id INTEGER UNIQUE`; migration uses a partial unique index `WHERE musicfolder_id IS NOT NULL`. Functionally equivalent today, but fresh and upgraded DBs hold different index objects in `sqlite_master`. Pick one shape.
>
> ### Minor
> - **`nextFolderId()` not transactional.** `MAX(musicfolder_id)+1` and the INSERT are separate statements. Safe at boot but fragile if peer insertion ever moves to admin runtime — wrap in `db.transaction(...)` or comment the assumption.
> - **`parseInt(q.musicFolderId, 10)`** in `hub/src/routes/subsonic.ts:489` has no NaN guard; falls through to `WHERE = NaN` → empty list. Works, but an explicit `Number.isFinite` would be clearer.
> - **SPA folder slug** `parseInt(slug.slice("folder-".length), 10)` accepts `folder-7xyz` as folder 7. `Number.isFinite` doesn't catch this — use `/^folder-\d+$/` or compare back to `String(id)`.
> - **AlbumsPage title flicker.** `title: folder ? peerDisplayName(folder.name) : "Loading…"` flashes "Loading…" on cold navigation. Cosmetic.
>
> ### Nits
> - Subsonic compat test imports `os` inside the function — move to top of file.
> - Legacy `instanceId` alias has no deprecation timeline; open a follow-up issue.
>
> ## Risk
> - **Migration:** low. Additive column + partial unique index, idempotent. The SQLite `ADD COLUMN ... UNIQUE` bug from this morning was caught and fixed in 8ca8949.
> - **Compat:** low. Legacy alias preserved; spec-compliant int ids returned.
> - **Perf:** negligible — one indexed lookup per filtered `getAlbumList2`.
>
> ## Verdict
> Ship after addressing the **peer-name fallback** (medium) — that one is a visible UX regression for operators who register a peer without an explicit display name. Other items are follow-ups.

PR #128, opened 2026-04-29
Per-user starred tracks/albums/artists + Favorites views (#104)

### LINKED ISSUE #104: Support for "star" or favorite tracks
Part of full Subsonic implementation ( #70 )

Use Subsonic features to set and query "starred" files. Star status should be per-user. Just because I "love" an album doesn't mean other people streaming from my instance will also love that album. Because a user might want to favorite a track that is from another hub, the tracking of user favorites needs to be done in Poutine, and not passed through to Navidrome.

Implementation should also add a new Albums view (similar to #82 ) for "Favorites".

How should favorites of individual tracks (instead of whole albums) be handed? The favorite mark could be set on the tracks as displayed under an Album view. In order to view what tracks are starred / favorites, add a new Nav heading for "Playlists" and create a virtual "Favorites" playlist under that, which contains all tracks that the user

benders:
## Summary

- New `user_stars` table; single shape covers tracks, albums, and artists. Keyed on `(user_id, kind, target_id)`. ON DELETE CASCADE on the user. Stars are local to the hub the user logs into; not federated.
- Subsonic endpoints: `star`, `unstar`, `getStarred`, `getStarred2`, plus `getAlbumList2?type=starred` (per spec). `id`/`albumId`/`artistId` all accepted; kind classified by id prefix. Existing builders (`getAlbum`, `getArtist`, `getAlbumList2`, `getSong`, `search3`) emit a `starred` ISO 8601 timestamp on entities the user has starred.
- SPA: shared `<StarButton>` mounted on every track row, album header, and artist header (optimistic via React Query). New `/library/favorites` view (uses `getAlbumList2?type=starred`). New `Playlists` sidebar group with a virtual `Favorites` playlist (`/playlists/favorites`) backed by `getStarred2`.
- Docs: `docs/opensubsonic.md` updated; new `user_stars` paragraph in `docs/hub-internals.md`.

Closes #104.

## Phased commits

| # | Commit  | Scope                                                              |
|---|---------|--------------------------------------------------------------------|
| 1 | e49445c | DB schema + idempotent migration                                   |
| 2 | e3b8cb2 | star/unstar/getStarred/getStarred2 endpoints                       |
| 3 | 6a18a04 | `getAlbumList2?type=starred` + `starred` annotation on existing builders |
| 4 | d590c85 | SPA: Albums Favorites view + StarButton on tracks/albums/artists   |
| 5 | 4cc0929 | SPA: Playlists page with virtual Favorites playlist                |
| 6 | 79cbc52 | Docs                                                               |

## Test plan

- [x] `pnpm typecheck`
- [x] `pnpm test` — 324 hub + 50 frontend passing
- [x] `pnpm test:federation` — sanity check (no per-user state federated; expected to pass unchanged)
- [x] `pnpm test:compat` — new cases `test_star_unstar_round_trip` and `test_get_album_list2_starred`
- [x] Manual: log in to running hub, star a track, confirm: appears in Playlists/Favorites, album shows yellow star in header, getAlbumList2 type=starred contains the album, second user does NOT see the same stars
- [x] Manual: 3rd-party Subsonic client (DSub / Submariner) round-trip via `getStarred2`


benders:
@claude review of PR #128 (per-user starred + Favorites views)

> ## Overview
> Adds a `user_stars` table keyed on `(user_id, kind, target_id)`, wires Subsonic `star`/`unstar`/`getStarred`/`getStarred2`/`getAlbumList2?type=starred`, annotates existing builders with an ISO `starred` timestamp, and ships an SPA `<StarButton>` plus `/library/favorites` and a virtual `Playlists/Favorites` page. Solid scope discipline — local-only, no federation, orphan-tolerant.

> ## What's good
> - Schema choice is clean: single table + composite PK + `idx_user_stars_lookup(user_id, kind, starred_at DESC)` lines up with the read paths in `buildStarredEnvelope`.
> - Intentional non-FK with read-time JOIN filtering is the right call for a sync-target world; the orphan test (`hub/test/star-routes.test.ts`) covers it.
> - `INSERT OR IGNORE` / `DELETE` make star/unstar idempotent; per-user scoping verified via the `alice` test.
> - `getAlbumList2?type=starred` reuses the existing query shape via `EXISTS` — minimal blast radius on a hot path.
> - Test coverage is appropriate: round-trip, prefix classification per kind, orphan filtering, per-user scoping, legacy `getStarred` envelope shape, and `subsonic-compat` updated.

> ## Concerns
>
> **1. `classifyStarId` "t" prefix is too permissive.**
> `hub/src/routes/subsonic.ts` checks `encoded.startsWith("t")` after `"al"`/`"ar"`. Anything that doesn't start with `al`/`ar` and starts with `t` is treated as a track and `slice(1)` is taken as the raw UUID. In practice ids are always server-encoded, but a malformed `id=tomato` would silently insert `omato` as a track target. Consider matching the actual encoding pattern (e.g. `t` followed by a hex/uuid shape) and returning `null` otherwise — same effect on legitimate clients, defensive against junk.
>
> **2. `collectStarTargets` quietly accepts unprefixed `albumId`/`artistId`.**
> `id.startsWith("al") ? slice(2) : id` means a raw UUID passed as `albumId=<uuid>` gets stored verbatim, which happens to work because the JOIN target is the unified UUID. Worth a one-liner comment so a future reader doesn't "fix" it.
>
> **3. `<StarButton>` is not actually optimistic.**
> PR description says "optimistic via React Query" but the component uses `onSuccess` to invalidate — the user sees a roundtrip flicker before the icon flips. Either drop the word "optimistic" from the docstring, or wire up `onMutate` to set the local query cache and `onError` to roll back. Not blocking, but the gap between description and code will confuse the next reader.
>
> **4. `getStarred` envelope shape.**
> The legacy `getStarred` returns the `getStarred2` envelope (`artist`/`album`/`song` with ID3 builders). Spec-strict `getStarred` returns `Child` shape under `directory`-style entries. Several real clients (DSub, Ultrasonic) tolerate the ID3 form, but worth flagging in `docs/opensubsonic.md` as a known divergence rather than just "same envelope under legacy key", in case a strict client trips on it later.
>
> **5. Songs query in `buildStarredEnvelope` has 4 correlated subqueries on `track_sources` per row.**
> Fine at typical "starred songs" cardinality but degrades on power users with thousands of stars. A single `JOIN` to a `(SELECT ... ORDER BY bitrate DESC LIMIT 1 PER unified_track_id)` window-function subquery would collapse them. Defer until someone actually has the cardinality to care.
>
> **6. `albumCount` for starred artists uses `LEFT JOIN unified_release_groups urg ON urg.artist_id = ua.id`.**
> This counts only release groups whose primary `artist_id` matches; collaborations / V/A appearances are missed. Matches the rest of the codebase's artist→album linkage so probably consistent, but worth confirming that's the same behaviour as `getArtist` returns elsewhere.
>
> **7. Minor: docstring on `StarButton` says "backend classifies by prefix"** — true for `id`, but not for `albumId`/`artistId` (caller already classified). Either drop the SPA's other id forms or document that the SPA only ever uses `id`.

> ## Nits
> - `toIsoStarred` is doing string surgery on a SQLite timestamp; if any other code path ever inserts an already-ISO value, the `includes("T")` check covers it. Consider centralizing this in a shared helper since this is the third place I've seen a SQLite→ISO conversion in the codebase.
> - `hub/src/routes/subsonic.ts:1163` — declaring the prepared statement inside the route handler means it's re-prepared on every request. Move to module-scope or memoize on `app`. Same applies to the unstar handler and the three queries in `buildStarredEnvelope`. better-sqlite3 caches plans on `prepare`, but allocation churn is avoidable.

> ## Risk
> Low. New table, idempotent writes, behind auth, no federation surface. Migration is `CREATE TABLE IF NOT EXISTS` so re-runs are safe. Rollback = drop the table.

> ## Verdict
> Approve with the "t" prefix tightening (concern #1) addressed before merge; the rest can land as follow-ups.

PR #132, opened 2026-04-30
fix(db): fix musicfolder_id migration for SQLite compatibility

loving-grace:
Fixes the 'no such column: musicfolder_id' error when upgrading from pre-0.4.4 versions.

## Root Cause

The `createDatabase()` function runs `db.exec(schema)` before `ensureColumns()`. On upgrade from pre-#123, the `CREATE TABLE IF NOT EXISTS instances` statement is a no-op (the table already exists), so `musicfolder_id` is never added. The very next statement in schema.sql — the partial unique index on that column — fails because the column doesn't exist yet, before the `ALTER TABLE` migration ever runs.

## Changes

- Move the `musicfolder_id` index creation from `schema.sql` to `ensureColumns()` after the `ALTER TABLE` adds the column
- Add logging to migration functions for better debugging
- Add comprehensive tests for the migration path (#123)

Note: The original `UPDATE ... FROM` backfill syntax was correct — SQLite has supported it since version 3.33 (Nov 2020).

### DIFF
--- a/hub/src/db/client.ts
+++ b/hub/src/db/client.ts
@@ -19,6 +19,10 @@ function dropLegacyTables(db: Database.Database): void {
  * Apply additive column migrations for existing databases.
  * SQLite supports ADD COLUMN idempotently via PRAGMA table_info check.
  */
+function logMigration(msg: string): void {
+  console.log(`[DB Migration] ${msg}`);
+}
+
 /**
  * Migrate users.password_hash (Argon2id) → users.password_enc (AES-256-GCM).
  *
@@ -35,11 +39,13 @@ function migrateUserPasswords(db: Database.Database): void {
   const names = new Set(cols.map((c) => c.name));
 
   if (!names.has("password_enc")) {
+    logMigration("Adding password_enc column to users table");
     db.exec(
       "ALTER TABLE users ADD COLUMN password_enc TEXT NOT NULL DEFAULT ''",
     );
   }
   if (names.has("password_hash")) {
+    logMigration("Dropping password_hash column from users table");
     db.exec("ALTER TABLE users DROP COLUMN password_hash");
   }
 }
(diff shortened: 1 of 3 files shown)


benders (changes requested):
@claude:

> **Root cause confirmed in commit 56686c8 (#123):** `createDatabase()` runs `db.exec(schema)` before `ensureColumns()`. On upgrade from pre-#123, `CREATE TABLE IF NOT EXISTS instances` is a no-op so `musicfolder_id` is never added, and the very next statement in `schema.sql` — the partial unique index on that column — fails before the `ALTER TABLE` migration ever runs.
>
> **Approve in spirit:** removing the index from `schema.sql` and moving it into `ensureColumns()` after the `ALTER TABLE` is the correct minimal fix. The regression test reproduces the upgrade path well.
>
> **Please change before merge:**
>
> 1. The PR description claims "UPDATE ... FROM is PostgreSQL-only." That is incorrect — SQLite has supported `UPDATE ... FROM` since 3.33 (Nov 2020), and better-sqlite3 11 ships a newer SQLite. The original `UPDATE ... FROM` backfill was not the bug. Please revert that rewrite (the new correlated-subquery form with `WHERE instances.id = id` relying on the inner unqualified `id` is harder to read) and fix the PR description.
>
> 2. Drop the `try { ... } catch (err) { console.error(...); throw err; }` wrappers around each migration function. better-sqlite3 already throws with the failing SQL; the wrappers add noise without information. Keep the `logMigration(...)` lines — those are useful.
>
> Once those two are addressed this is good to merge.

loving-grace:
I've addressed both points:\n\n1. Reverted the UPDATE syntax to the original  form (which is correct for SQLite 3.33+)\n2. Removed all the  wrappers while keeping the  calls\n\nThe PR description has also been updated to correctly explain the root cause (index being created before the column existed, not the UPDATE syntax).

loving-grace:
I've addressed both points:

1. Reverted the UPDATE syntax to the original `UPDATE ... FROM` form (which is correct for SQLite 3.33+)
2. Removed all the `try/catch` wrappers while keeping the `logMigration()` calls

The PR description has also been updated to correctly explain the root cause (index being created before the column existed, not the UPDATE syntax).

gnarg:
Not that we strictly need it, but this worked for me locally, FWIW.

benders (approved):
@claude:

> Both requested changes addressed:
>
> - `UPDATE ... FROM` backfill restored to the original syntax; PR description corrected.
> - try/catch wrappers dropped; `logMigration(...)` calls retained where they add signal.
>
> Schema/index move + regression test look good. LGTM.

Issue #154, opened 2026-05-09
UI: distinguish 'Last Seen' vs 'Last Sync' for peers


benders:
From #14 description. Peer info box should show:
- Last Seen: most recent successful contact (any kind, including /api/health poll)
- Last Sync: most recent successful library sync
Currently only last_synced_at + last_seen exist on instances; ensure auto-sync's /api/health poll updates last_seen, sync updates last_synced_at, and the SPA renders both.

Issue #153, opened 2026-05-09
Gossip rate limiting + transitive trust policies


benders:
Follow-up to #147. Defensive hardening:
- Cap gossip ingest per sync round (avoid flood from compromised peer)
- Optional admin policy: only auto-accept peers gossiped by N hubs, or only peers within K hops of a manually-trusted root
- Backoff on repeated invalid invitations from the same source

Issue #152, opened 2026-05-09
Peer record backup/restore tooling


benders:
Follow-up to #147. Replaces the peers.yaml workflow which is removed in v0.5.0.
- Export peer rows + invitation provenance to a portable file
- Import on a fresh hub (e.g. disaster recovery, instance migration)
- Validate signatures on import; reject any peers whose inviter key doesn't verify
- CLI subcommand or admin endpoint

Issue #151, opened 2026-05-09
Admin enable/disable for discovered peers


benders:
Follow-up to #147. Discovered (gossiped) peers are inserted enabled by default. Admin needs:
- Per-peer enable/disable toggle in UI
- Disabled peers: skipped during sync, hidden from federation outbound, but row + provenance retained
- Surface 'discovered via <peer>' provenance

PR #155, opened 2026-05-09
feat: peer gossip + signed invitations (federation v5, target 0.5.0)

### LINKED ISSUE #14: Automatically keep peers synced

benders:
## Summary

Implements #147 — peer gossip + signed invitations — and bundles #14 (auto peer sync) since gossip rides on the sync transport.

Federation API bumps to **v5** (skipping v4). App targets **0.5.0**.

## All five phases landed

- [x] **Phase 1** — Schema + DB-backed peer registry. `peers.yaml` removed (hard cut).
- [x] **Phase 2** — Invitation issue + acceptance protocol (`/admin/peers/invite`, `/federation/handshake`, `/admin/peers/accept`).
- [x] **Phase 3** — Gossip transport on sync (`/federation/peers`, ingest in `sync.ts`).
- [x] **Phase 4** — Periodic auto-sync (closes #14).
- [x] **Phase 5** — Admin UI + version bumps + docs.

## Hard-cut migration

`peers.yaml` is no longer read. Legacy peer rows from v0.4.x DBs have no public_key and are removed automatically on first boot. Operators re-invite peers via the new admin UI flow.

## Trust model

1. Inviter's Ed25519 signature on the canonical invitation payload (verifiable by anyone who has the inviter's public key).
2. Invitee's Ed25519 signature over the invitation nonce — proves ownership of the claimed pubkey, defends against a leaked invite being redeemed by a third party.
3. `/api/health` reachability + `apiVersion >= 5` check.
4. Single-use enforcement via the new `invitations` table.

Gossiped entries carry the original signed invitation; receivers verify it against the named inviter's public key without prior trust.

## Follow-ups (filed as separate issues)

- #150 tombstones / removal propagation
- #151 admin enable/disable for discovered peers
- #152 peer record backup/restore tooling
- #153 gossip rate limiting + transitive trust policies
- #154 'Last Seen' vs 'Last Sync' UI distinction

## Test plan

- [x] `pnpm typecheck` clean (hub + frontend)
- [x] `pnpm test` — 386/386 hub, 62/62 frontend
- [x] `pnpm test:federation` integration scenario (rewritten for v5 invite/accept; gossip-discovers third hub from second hub)
- [ ] Manual on a live test instance: generate invite, accept on second instance, verify peer appears + auto-sync works


benders:
FROM @claude:

> ## Review — federation v5 (peer gossip + signed invitations)
>
> Code quality is high — clean separation (`invitations.ts`, `gossip.ts`, route handlers), tests for each new module, docs updated alongside. Findings below in order of importance.
>
> ### Correctness
>
> **MAJOR — Nonce single-use is racy.** `hub/src/routes/federation.ts` reads `invitations.consumed_at` (~L101–118), then `await fetch(${inviteeUrl}/api/health)` (~L150, ~5 s timeout), *then* runs the admit transaction (~L188–215). Two concurrent handshakes for the same invitation, but with different `invitee.id`s, can both pass the SELECT, both await health, and both run the txn — both `INSERT`s succeed (distinct PKs) and both `UPDATE`s run. Result: one invitation admits two peers. Fix: re-check inside the transaction, or change the consume to `UPDATE invitations SET consumed_at=…, consumed_by_id=? WHERE id=? AND consumed_at IS NULL` and abort the txn when `changes === 0`.
>
> **MINOR — invitation-table `expires_at` and the embedded payload's `expires_at` are independently checked.** They're set together at issue time but it's worth a comment that the row check is the authoritative one (or drop it and rely on the signed payload's value — single source of truth).
>
> ### Security
>
> **MAJOR — Gossip trust is non-transitive but treated as transitive.** `gossip.ts` admits any peer entry whose embedded invitation verifies against its own `inviter_public_key`. There is no requirement that `inviter_id` be a peer we already trust. An attacker who is *any* admitted peer (or who lures us into accepting *one* of their invitations) can gossip arbitrary `{inviter_id: "anything", inviter_public_key: <attacker key>, signature: <valid>}` entries and we will admit them. Follow-up #153 is filed, but at minimum land a check now: `inviter_id` must already exist in `instances` (i.e. the inviter is a peer we already trust). Without that, an admitted peer can vouch in arbitrary "inviters" and grow the cluster unilaterally.
>
> **MAJOR — SSRF via `/federation/handshake`.** L150 `fetch(\`${inviteeUrl}/api/health\`)` runs against any URL the invitee submits. For *targeted* invites the inviter chose the URL (low risk), but for *open invites* (`invitation_url = null`) any holder of the invitation can make the inviter hub GET an arbitrary URL — useful for internal-host probing (response code, timing, JSON shape leak via the 502 error message which embeds `String(err)`). At minimum: block private IP ranges and non-http(s) schemes; don't echo `String(err)` back in the 502.
>
> **MINOR — Open invites under-warned.** `invitations.ts:25` notes "null = open invite (any URL)" but the admin UI / README should explicitly warn: a leaked open invite admits *any* hub that can post a valid pubkey-proof. Consider defaulting `invitee_url` to required in the admin form and making "open" an explicit checkbox.
>
> **NIT — Health-check has no max body size.** A malicious invitee can return a huge JSON; `res.json()` will buffer it. Cap with a size-limited reader if you want to harden.
>
> ### Quality / robustness
>
> **MINOR — Auto-sync timers not `unref()`'d.** `auto-sync.ts:47, 68`: if shutdown skips `stop()`, the process won't exit. Add `.unref()` after each timer creation.
>
> **MINOR — `gossip.ts` recomputes `MAX(musicfolder_id)+1` per insert inside a loop without an outer transaction.** Not a correctness bug (better-sqlite3 is sync), but wrapping the loop in a `db.transaction()` is faster and gives all-or-nothing semantics if you later batch.
>
> **MINOR — Legacy-peer prune is silent on operator action.** `hub/src/db/client.ts:100–112` deletes rows with `public_key IS NULL` and logs a single warn line. Consider also writing an entry to the activity log so operators see it in the UI — otherwise a v0.4.x → v0.5.0 upgrade silently drops federation until re-invites are issued.
>
> ### Tests
>
> Strong coverage on happy paths. Gaps to consider:
> - Concurrent handshake on same nonce (the race above).
> - Handshake `inviteeUrl` pointing at a private/loopback host.
> - Gossip entry whose `inviter_id` is unknown to the receiver.
> - Migration test that legacy v0.4.x rows are pruned and the admin sees a clear signal.
>
> ### Verified-OK
>
> - `/admin/peers/invite` and `/admin/peers/accept` are both behind `requireOwner` (admin-only). ✓
> - `/federation/peers` is behind `requirePeerAuth`. ✓
> - Gossip rejects entries whose payload `inviter_*` fields disagree with the entry's claimed `inviter_*` (gossip.ts:99–113). ✓
> - Invitee proof-of-key (`verifyInviteeProof`) is checked before admit. ✓
> - `inviter_id` and `inviter_public_key` collision against self are checked at handshake. ✓
>
> ### Recommended pre-merge fixes (in order)
>
> 1. Make nonce consumption atomic.
> 2. Require gossip `inviter_id` already exist in `instances`.
> 3. SSRF guard on the handshake `inviteeUrl` fetch + redact `String(err)` from the 502 body.
> 4. `.unref()` the auto-sync timers.
>
> Will follow up with commits addressing 1–4.

Issue #108, opened 2026-04-26
Ability to stream to local Sonos devices directly


benders:
What would it take to be able to stream audio from the web interface directly to Sonos devices?

benders:
@claude:
> Research notes — no implementation drafted, correct path unclear.
>
> **Three possible approaches:**
>
> **1. Local UPnP/SOAP (lightest)**
> Sonos speakers expose UPnP \`AVTransport\` on LAN port 1400. Send \`SetAVTransportURI(<url>)\` + \`Play\`; speaker pulls audio directly from hub. No Sonos cloud account, no app registration.
> - Libs: \`sonos\` (node-sonos), \`@svrooij/sonos\`.
> - Discovery via SSDP multicast (\`urn:schemas-upnp-org:device:ZonePlayer:1\`) — must run on hub host network, not browser.
> - Works only when hub + speakers share a LAN.
>
> **2. Sonos Cloud Control API**
> OAuth2 via Sonos developer portal. \`loadAudioClip\` for short clips; full playback still requires URLs served by a registered SMAPI service. Not useful standalone.
>
> **3. SMAPI music service (heaviest, best UX)**
> Register Poutine as a Sonos music service. SOAP server implementing \`getMetadata\`, \`getMediaURI\`, \`search\`, etc. Users add \"Poutine\" in the Sonos app and browse natively. Multi-week scope, requires Sonos developer approval.
>
> **Open problems regardless of path:**
> - **Stream auth** — Sonos fetches URLs with no headers. Current \`/rest/stream\` needs Subsonic creds in querystring (leaks long-lived token). Want short-lived signed URLs (HMAC + exp).
> - **Federation** — Sonos can't auth across hubs. Either restrict casting to local-library tracks, or have hub proxy peer streams behind a local signed URL.
> - **Queue semantics** — \`SetAVTransportURI\` plays one track. Real queue support needs \`AddURIToQueue\` + \`x-rincon-queue:<UUID>#0\`.
> - **Discovery surface** — SSDP from hub means new endpoints (\`GET /sonos/devices\`, \`POST /sonos/play\`) and a device cache.
> - **Formats** — Sonos S2 handles MP3/AAC/FLAC/ALAC/WAV/OGG. Existing MP3 transcode pipeline is fine.
>
> **Format/UX questions still open:**
> - Should casting replace the local web player or run alongside?
> - Group/zone handling (stereo pairs, bonded rooms)?
[...]

PR #162, opened 2026-05-11
feat(sonos): stream to local Sonos devices via UPnP/SOAP (refs #108)

benders:
## Summary

- Adds an opt-in **Sonos casting** sink for the bottom-of-screen player. When enabled, playback routes to a Sonos device on the LAN instead of the browser.
- Discovery via **SSDP M-SEARCH**; control via **SOAP** (AVTransport + RenderingControl) on each device's `:1400`.
- Sonos fetches streams from a new **`/cast/stream/:trackId`** endpoint authenticated by short-lived HMAC tokens (no Subsonic credential leakage to the device).

## Scope

**Feature-flagged.** Off by default. Enable with:

```
SONOS_ENABLED=true
POUTINE_LAN_URL=http://<lan-ip>:3000
docker compose -f docker-compose.yml -f docker-compose.sonos.yml up -d
```

Host networking is required so SSDP multicast (UDP 1900) works.

## Decisions

- **Queue model:** app-managed, one track at a time. Shuffle/repeat live in the player store; backend pushes the current track and the frontend advances via `next()` when the state poll sees `PLAYING → STOPPED`.
- **Stream auth:** HMAC-signed tokens (TTL 1h), secret derived from the instance Ed25519 key. No new on-disk secret.
- **LAN URL:** new `POUTINE_LAN_URL` env var, required when Sonos enabled.
- **Device persistence:** none — defaults to local browser playback each session.

## Files

| Layer    | New                                                                                                                                                                                   | Modified                                                                                                                                |
|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
| Backend  | `hub/src/services/sonos-discovery.ts`, `sonos-control.ts`, `cast-tokens.ts`; `hub/src/routes/sonos.ts`, `cast.ts`                                                                     | `hub/src/config.ts`, `server.ts`, `services/stream-tracking.ts`                                                                          |
| Frontend | `frontend/src/components/player/DevicePicker.tsx`                                                                                                                                     | `PlayerBar.tsx`, `stores/player.ts`, `lib/api.ts`                                                                                       |
| Infra    | `docker-compose.sonos.yml`                                                                                                                                                            | `docker-compose.yml`, `example.env`                                                                                                     |
| Docs     | —                                                                                                                                                                                     | `docs/hub-internals.md`, `docs/system-architecture.md`, `README.md`                                                                     |
| Tests    | `sonos-discovery.test.ts`, `sonos-control.test.ts`, `cast-tokens.test.ts`                                                                                                             | —                                                                                                                                       |

## Test plan

- [x] `pnpm test` — 39 hub files / 402 tests + 12 frontend files / 62 tests passing
- [x] `pnpm typecheck` clean
- [x] 15 new unit tests: SSDP parsing, device-description XML, DIDL-Lite, SOAP envelopes, HMAC token sign/verify round-trip, expiry rejection, cross-track rejection
- [x] Manual e2e (requires host networking + real Sonos on LAN):
  - [x] `SONOS_ENABLED=true` boot logs "Sonos discovery started" + "Sonos casting enabled"
  - [x] `curl /api/sonos/devices` lists at least one device with `room`
  - [x] Click cast icon in PlayerBar → device list populates → select device
  - [ ] Press play → audio plays from the Sonos device; transport state polled every 1.5s
  - [ ] Pause / resume / seek / volume each reflect on device
  - [x] End-of-track auto-advances to next queue entry
  - [ ] Switch sink back to "This browser" → playback resumes locally
  - [ ] `SONOS_ENABLED=false` → no cast icon; only local playback
- [ ] Edge cases:
  - [ ] Cast token reuse on a different track ID rejected (verified by unit test)
  - [ ] Expired token rejected (verified by unit test)
  - [ ] Device disappearing mid-stream (state poll fails silently, UI keeps last position)


benders:
FROM @claude:

> ### Review — generally clean, but a few issues to address before un-drafting

> **🔴 Blocker — \`docker-compose.sonos.yml\` breaks Navidrome reachability.**
>
> With \`network_mode: host\`, the hub leaves Docker's bridge network and loses access to the embedded DNS that resolves \`navidrome\` (compose-defined service name). The base file still sets \`NAVIDROME_URL=http://navidrome:4533\` and navidrome is on an \`internal: true\` network with no published ports — so an operator who follows the README instructions will get Sonos discovery working, but local playback will fail entirely because the hub can't reach Navidrome.
>
> Options to fix:
> 1. In \`docker-compose.sonos.yml\`, also publish a port on navidrome (e.g. \`127.0.0.1:4533:4533\`) and override \`NAVIDROME_URL=http://127.0.0.1:4533\`.
> 2. Document the trade-off prominently and require operators to set \`NAVIDROME_URL\` themselves.
>
> Suggest option 1, in the override file itself, so the documented one-liner Just Works.

> **🟠 \`cast-tokens.ts\` — the \`userId\` parameter is a lie.**
>
> \`signCastToken\` accepts \`userId\` and includes it in the signed message, but \`verifyCastToken\` hardcodes \`userId = \"cast\"\` on the verify side (see comment block at hub/src/services/cast-tokens.ts:55-58). So signing with anything other than \`\"cast\"\` produces a token that *won't verify against itself*. Current call sites all pass \`\"cast\"\` so it works today, but the API is a footgun.
>
> Either:
> - Drop \`userId\` from the public signing API entirely (it's not load-bearing yet), or
> - Encode userId into the token (\`<sig>.<exp>.<userId>\`) so verify can recover it and audit per-user casts.

> **🟠 Switching to Sonos auto-plays even if local was paused.**
>
> The track-change effect at \`PlayerBar.tsx:175\` calls \`sonosPlay(...)\` whenever \`isSonos\` flips true and a track is loaded, regardless of \`isPlaying\`. So a user who pauses local audio, opens the picker, and selects a Sonos device will get unexpected playback on their living-room speaker. Suggest gating on \`isPlaying\` in that effect (and pushing track URI without auto-Play if not). Backend \`POST /api/sonos/devices/:id/play\` could take an explicit \`{ autoplay: boolean }\` toggle, or split the route into \`load\` + \`play\`.

> **🟡 State poll runs without coalescing.**
>
> \`window.setInterval(() => void tick(), 1500)\` fires every 1.5s regardless of whether the previous SOAP \`GetTransportInfo + GetPositionInfo\` round-trip finished. With a 5s timeout in \`sonos-control.ts\` and a slow LAN, requests can pile up. Easy fix: track an \`inFlight\` flag and skip ticks while one is outstanding.

> **🟡 \`sink\` in effect deps is an object literal.**
>
> Several PlayerBar effects use \`[isSonos, sink, currentTrack?.id, ...]\` as their deps. \`sink\` is a new object every time \`setSink\` is called (zustand creates new state objects), so calling \`setSink\` with semantically identical values re-fires \`sonosPlay\` and rebuilds the state-poll interval. In practice the user only calls \`setSink\` from the picker so it's mostly fine, but it would be safer (and cheaper) to depend on \`sink === \"local\" ? null : sink.deviceId\` directly.

> **🟡 No DevicePicker in the empty-state branch.**
>
> When there's no current track, \`PlayerBar.tsx:219-226\` renders the \"No track playing\" placeholder with no DevicePicker. A user who just loaded the app can't pre-select a Sonos device before queueing a track. Move the picker into the always-rendered shell.

> **🟢 Test coverage is good for the pure helpers but thin for routes.**
>
> 15 new unit tests cover SSDP parsing, DIDL-Lite, SOAP envelopes, token sign/verify — nicely focused. Route handlers (\`routes/sonos.ts\`, \`routes/cast.ts\`) have no tests. Pattern in this repo (\`hub/test/stream.test.ts\`, \`subsonic-routes.test.ts\`) is to use Fastify \`inject\` with a fake upstream HTTP server. At minimum, add tests for:
> - \`/cast/stream/:trackId\` returning 401 (no token), 403 (bad token), 404 (no track)
> - \`/api/sonos/devices/:id/play\` returning 500 when \`POUTINE_LAN_URL\` is unset
> - Token sign/verify integration via the route, not just the unit (catches the \"userId\" footgun)

> **🟢 Nits**
> - \`hub/src/services/sonos-control.ts:74\` — \`xmlEscape\` is applied to the streamUri inside \`<res>\`. Sonos expects the URL XML-decoded, which this gives it — correct, but worth a comment because it's load-bearing.
> - \`hub/src/services/sonos-discovery.ts:160\` — \`setMulticastTTL(2)\` is fine for LAN, but most SSDP clients set it to 4. Doesn't matter for the typical case.
> - The frontend \`apiFetch\` redirects to \`/login\` on 401. If a Sonos device disappears mid-session and the state poll happens to hit a 401 (shouldn't, but) the user gets bounced to login. Minor.
> - \`hub/src/server.ts:308\` reuses the federation private key DER to derive the cast secret. Solid choice — no new on-disk material — but worth noting in the security section of \`docs/hub-internals.md\` that compromise of the federation key compromises cast tokens (the inverse already being true).

> ### Verdict
>
> Architecture and code structure are right. Two correctness issues (compose override / token userId) and one UX issue (auto-play on sink switch) should block merge; the rest are nice-to-haves that can land in a follow-up.

benders:
FROM @claude:

> ### Review fixes pushed (commit 18ea4ed)
>
> | Finding                                              | Status | Notes                                                                                                                |
> |------------------------------------------------------|--------|----------------------------------------------------------------------------------------------------------------------|
> | 🔴 Compose override breaks Navidrome reachability   | ✅     | Override now publishes navidrome on \`127.0.0.1:4533\` + sets \`NAVIDROME_URL=http://127.0.0.1:4533\` on the hub      |
> | 🟠 \`cast-tokens\` \`userId\` footgun                | ✅     | Dropped from API. Sign/verify now agree on \`<trackId>\|<exp>\`. \`verifyCastToken\` returns boolean                  |
> | 🟠 Auto-play on sink switch                          | ✅     | New \`autoplay\` flag on \`/play\` (default true, backwards compatible). Frontend passes \`autoplay: isPlaying\`     |
> | 🟡 State poll coalescing                             | ✅     | \`inFlight\` flag in the polling effect                                                                              |
> | 🟡 \`sink\` object identity in deps                  | ✅     | Effects pin to a \`deviceId: string \| null\` derived from sink                                                       |
> | 🟡 DevicePicker missing in empty-state               | ✅     | Picker now renders in the no-track placeholder                                                                       |
>
> Tests: 403 hub / 62 frontend passing; \`pnpm typecheck\` clean. Cast-tokens test suite updated for the simplified API (7 cases, including malformed-token coverage).
>
> Green review items (route-handler tests, multicast TTL, etc.) left for a follow-up — flagging so they don't get lost.

benders:
FROM @claude:

> ### Auth + cast-token attribution fix (commit 616b33a)
>
> **1. JWT auth on `/api/sonos/*`.** Added `requireAuth` preHandler to the whole route plugin. Previously unauthenticated — under host networking (the only mode where Sonos works), any LAN-reachable client could enumerate rooms and trigger playback. Now matches the `/admin/*` pattern.
>
> **2. Username encoded in cast tokens.** Token wire format is now `<sig>.<exp>.<base64url(username)>`; sign and verify agree on the message `trackId|exp|username`. \`/api/sonos/play\` looks up the JWT user, signs a token carrying their username, and \`/cast/stream\` uses the recovered name for stream-tracking attribution and federated \`asUser\` (instead of the hardcoded \"cast\" sentinel from the previous revision).
>
> **Tests:** cast-tokens suite expanded to 9 cases — username-segment tampering, unicode/special-char usernames, 3-segment malformed input. 405 hub / 62 frontend passing; typecheck clean.

benders:
FROM @claude:

> ## Validation finding: Docker Desktop on macOS cannot deliver SSDP multicast to containers
>
> Tested the `docker-compose.sonos.yml` override on macOS (Docker Desktop with experimental host networking enabled). Hub starts cleanly, Sonos discovery service initializes, but `GET /api/sonos/devices` always returns `{"devices": []}`.
>
> ### Diagnosis
>
> - Mac host directly: SSDP M-SEARCH returns 4 Sonos units on the LAN. ✅
> - Inside the hub container (host networking enabled): same M-SEARCH returns 0 responses. ❌
>
> Docker Desktop's "host networking" on macOS is implemented as a userspace VPN, not a true network-namespace share. UDP multicast (SSDP 239.255.255.250:1900) does not traverse it. This is a Docker Desktop limitation, not a Poutine bug.
>
> ### Reviewed mitigations
>
> Checked the techniques in [oneuptime's multicast-in-Docker guide](https://oneuptime.com/blog/post/2026-02-08-how-to-configure-docker-containers-for-multicast-traffic/view):
> 1. Disable IGMP snooping on bridge — Linux only
> 2. Macvlan networking — Linux only
> 3. Host networking — Linux only (the option this PR uses; works on Linux)
>
> None resolve the Mac Docker Desktop case.
>
> ### Other validation fixes folded into this PR during testing
>
> - `docker-compose.sonos.yml` was missing `networks: !reset null` on the hub service — Compose merges the base file's `networks:` list with `network_mode: host`, which Compose rejects as mutually exclusive. Fixed.
> - Added `extra_hosts: ["host.docker.internal:host-gateway"]` so the override is portable to Linux hosts (Docker Desktop auto-injects this; Linux doesn't).
> - Navidrome had to also move to `network_mode: host` (bound to `ND_ADDRESS=127.0.0.1`). On Docker Desktop, host-net containers cannot reach the Mac's loopback-published bridge ports, so the prior "publish on 127.0.0.1:4533" hop is broken on Mac.
>
> ### Recommendation
>
> Code in this PR is correct. Ship it. Document that **Sonos casting requires a Linux host** in the operator docs; on macOS, run the hub natively (outside Docker) if you want to test Sonos discovery against a real LAN.
>
> Filed [#178](https://github.com/benders/poutine/issues/178) (health-check should surface Navidrome reachability — masked the proxy failure during this validation) and [#179](https://github.com/benders/poutine/issues/179) (consider merging hub + Navidrome into one container — would eliminate this networking dance) as follow-ups.

benders:
FROM @claude:

> ## Validation fixes pushed (commit f0520bc)
>
> After running the hub natively on macOS (the only way to get SSDP multicast to work — Docker Desktop's host networking is a userspace VPN, see prior comment), casting end-to-end revealed three real bugs:
>
> ### 1. SQL join referenced a non-existent column
>
> Play handler queried \`unified_tracks.rg_id\`. Actual schema requires \`unified_tracks.release_id → unified_releases.release_group_id → unified_release_groups\`. Every cast 500'd with \`SQLITE_ERROR: no such column: ut.rg_id\`.
>
> ### 2. DIDL mime / stream content-type mismatch
>
> DIDL metadata hardcoded \`audio/mpeg\` but the cast URL didn't force a transcode, so FLAC sources were streamed as \`audio/flac\`. Sonos accepted the AVTransport URI (\`duration\` populated, volume control worked) but immediately landed in STOPPED. Fix: append \`?format=mp3\` to the cast URL. Lossless pass-through tracked separately in #180.
>
> ### 3. Compose override networking
>
> - \`network_mode: host\` collided with the base file's inherited \`networks:\` list (Compose merges keys; the two are mutually exclusive). Added \`networks: !reset null\`.
> - Hub on host net couldn't reach Navidrome's loopback-published port on Docker Desktop. Moved Navidrome to host net too, bound to \`ND_ADDRESS=127.0.0.1\`.
> - Added \`extra_hosts: host.docker.internal:host-gateway\` for Linux portability (Docker Desktop auto-injects; Linux doesn't).
>
> ### Test coverage
>
> Added \`hub/test/sonos-routes.test.ts\` exercising the play route through Fastify inject with stub discovery + control. Covers the two-hop join, the \`format=mp3\` query parameter, DIDL field population, 404 paths, and the auth gate. This file didn't exist before; the original PR had unit tests for \`sonos-control\` and \`sonos-discovery\` but no HTTP-route integration test, which is how the \`rg_id\` reference got past CI.
>
> Full suite: 40 files / 409 hub tests + 12 / 62 frontend tests, all green. Typecheck clean.
>
> ### End-to-end validation
>
> Ran hub natively on Mac against Dockerized Navidrome, cast \"Dancing Queen\" to Living Room (Sonos Era 100). State went \`STOPPED → TRANSITIONING → PLAYING\`, position advanced, audio audible, \`stop\` works. ✅
>
> Docs updated in \`docs/hub-internals.md\` to record the dual-host-net requirement and the Mac/Docker-Desktop multicast limitation.

benders:
FROM @claude:

> ## Bug 4: SPA-initiated casts 404'd on track lookup (commit e17f945)
>
> Validating from the SPA itself surfaced another mismatch. \`SubsonicSong.id\` in the frontend is the Subsonic \`remote_id\` (e.g. \`7jwQomahCwKbSjrAxtelmw\`), not the unified_tracks UUID. The play route queried unified_tracks by that string and returned \`{"error":"Track not found"}\` for every SPA cast.
>
> ### Fix
>
> Resolve via \`track_sources → instance_tracks\` first, fall back to treating the input as already-unified so direct API callers using the canonical UUID still work. The resolved unified id is used in the cast token and URL, since \`cast.ts /stream\` verifies the token against unified_tracks.
>
> ### End-to-end
>
> SPA → cast Come Together → Living Room. State went \`STOPPED → TRANSITIONING → PLAYING\`, position advanced, audio played (apparently quite loudly 🎸). ✅
>
> Test coverage extended (\`sonos-routes.test.ts\` now seeds the instance_tracks + track_sources chain and asserts the resolved cast URL contains the unified UUID, not the Subsonic id).

benders:
FROM @claude:

> ## Bugs 5 + 6 (commit 127db98)
>
> ### 5. SPA sends \`t<uuid>\` (encoded Subsonic song ids); play route 404'd
>
> The hub's \`/rest/getAlbum\` (and friends) return song IDs as \`encodeId("t", unified_tracks.id)\` — a "t" followed by the UUID. The earlier fix resolved bare unified UUIDs and Subsonic \`remote_id\` strings, but missed the "t" prefix. Every SPA-initiated cast except the proof-of-concept track 404'd.
>
> Play route now accepts all three id forms: \`t<uuid>\` (SPA), raw UUID (direct API), \`remote_id\` (older flows). Verified end-to-end with each.
>
> ### 6. Play/pause icon stuck while casting
>
> The Sonos polling effect in \`PlayerBar.tsx\` updated position + duration on each tick but never reflected the device's transport state back into the store. Any divergence — device-side pause, autoplay-after-SetAVTransportURI, cross-tab interaction — left the play/pause icon out of sync with the speaker. Now syncs \`isPlaying\` from \`s.state\` in the same tick. zustand no-ops on unchanged values, so this doesn't fight the existing toggle effect.
>
> ### Tests
>
> Added the t-prefix case to \`sonos-routes.test.ts\` (the most production-realistic of the three id forms — and the one PR #162 originally missed). All 6 sonos-routes tests green, full suite 411 hub + 62 frontend, typecheck clean.

Issue #175, opened 2026-05-13
DLNA MediaServer (ContentDirectory) for WMP / Xbox / generic UPnP browsers


benders:
> _FROM @claude:_
>
> Sub-issue of #102. **Prerequisite: #162 (Sonos casting) must land first** — it introduces the SSDP/SOAP/DIDL primitives this work will extract and reuse.
>
> ## Goal
>
> Let Windows Media Player Legacy (and Xbox, VLC, Kodi, BubbleUPnP, etc.) **discover and browse** the Poutine library as a UPnP MediaServer on the LAN. This is the direction #102 most likely actually wants: WMP is primarily a control point, not a render target.
>
> ## Scope
>
> Poutine advertises a UPnP \`MediaServer:1\` device and serves \`ContentDirectory:1\` Browse/Search SOAP requests. Clients fetch audio over HTTP from a Subsonic-credential-free stream endpoint (similar shape to the \`/cast/stream/:trackId\` route added in #162).
>
> ## New work (not reused from #162)
>
> - **SSDP advertiser** — periodic NOTIFY beacons (alive/byebye/update). #162 only does M-SEARCH. New code.
> - **Device description XML** at a stable URL (e.g. \`/dlna/device.xml\`) listing service types, control/event URLs, icons.
> - **ContentDirectory service handlers** — SOAP envelopes for \`Browse\`, \`Search\`, \`GetSortCapabilities\`, \`GetSearchCapabilities\`, \`GetSystemUpdateID\`. Returns DIDL-Lite container/item XML.
> - **Object hierarchy mapping** — translate Poutine's library (artists / albums / tracks / playlists) into a stable UPnP object ID tree. WMP caches IDs aggressively; they must be deterministic across restarts.
> - **Stream endpoint** sized for DLNA clients — must advertise proper \`Content-Type\`, \`Content-Length\` (or chunked), \`transferMode.dlna.org\`, \`contentFeatures.dlna.org\` headers. WMP rejects responses missing these.
> - **Transcoding decision** — WMP Legacy expects MP3/WMA. Decide: rely on source files being MP3, or wire a transcode path (likely out of scope for v1; document limitation).
> - **ConnectionManager:1** stub (required by spec; trivial).
>
> ## Reuse from #162
>
> - DIDL-Lite builder (once factored out into \`hub/src/services/didl.ts\`).
> - SOAP envelope helpers (once factored into \`soap.ts\`).
>
> ## Risk / open questions
>
[...]

PR #176, opened 2026-05-13
feat(dlna): MediaServer scaffold for WMP / Xbox / generic UPnP browsers (refs #175)

benders:
## Summary

Initial scaffold for the DLNA MediaServer (issue #175). Stacks on the Sonos casting branch (#162) — it factors the DIDL-Lite + SOAP helpers out of `sonos-control` into shared modules and reuses the host-networking compose override.

When `DLNA_ENABLED=true`, Poutine advertises itself as a UPnP `MediaServer:1` on the LAN. Windows Media Player, Xbox, Kodi, VLC, BubbleUPnP and other control points can browse the merged library and stream from `/dlna/stream/:trackId` (no Subsonic auth required — DLNA has no user identity).

## What's in scope

- **Shared primitives:** `services/didl.ts` and `services/soap.ts` extracted from `sonos-control.ts` so both casting and the DLNA server share the same helpers.
- **SSDP advertiser:** periodic NOTIFY `ssdp:alive`, `ssdp:byebye` on shutdown, M-SEARCH responder for `upnp:rootdevice` / `uuid:<udn>` / `MediaServer:1` / `ContentDirectory:1` / `ConnectionManager:1`.
- **Object hierarchy:** root → Music → {Artists, Albums, All Tracks} → drill-downs. Deterministic IDs over `unified_artists` / `unified_release_groups` / `unified_tracks`. The UDN itself is derived from `POUTINE_INSTANCE_ID` so it survives restarts (matters for WMP's aggressive client-side cache).
- **HTTP surface:** `device.xml`, ContentDirectory + ConnectionManager SCPDs, SOAP control endpoints (`Browse`, `GetSortCapabilities`, `GetSearchCapabilities`, `GetSystemUpdateID`, `GetProtocolInfo`, …), `/dlna/stream/:trackId` with `transferMode.dlna.org` + `contentFeatures.dlna.org` response headers.
- **Tests:** 30 new unit tests covering SOAP envelope/response/action parsing, SSDP packet construction, M-SEARCH parsing, target matching, object-ID parsing, and `Browse` against an in-memory schema-seeded DB.

## Authentication

DLNA has no notion of user identity. Two implications:

- `GET /dlna/stream/:trackId` is **unauthenticated** when `DLNA_ENABLED=true`. Anyone on the LAN can browse and stream. This is an explicit all-or-nothing decision — if that's unacceptable for an operator's network, leave the feature off.
- Stream activity is attributed to `DLNA_PSEUDO_USER` (defaults to the owner) for the activity feed and federated `asUser` routing.

## Compatibility notes

- v1 ships **pass-through only** — response MIME mirrors the source format (mp3 → `audio/mpeg`, flac → `audio/flac`, etc.). Windows Media Player Legacy reliably handles MP3/WMA only; everything else is the client's problem. Modern DLNA clients (Kodi, VLC, BubbleUPnP) handle FLAC and friends natively.
- `protocolInfo` 4th field advertises `DLNA.ORG_OP=01` (range support) + the standard streaming `DLNA.ORG_FLAGS` matrix. Expect a tuning pass against a real Windows box.

## Files

| Layer    | New                                                                                                                     | Modified                                                                          |
|----------|-------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------|
| Backend  | `hub/src/services/{soap,didl,ssdp-advertiser,dlna-objects}.ts`, `hub/src/routes/dlna.ts`                               | `hub/src/services/{sonos-control,stream-tracking}.ts`, `hub/src/{config,server}.ts` |
| Tests    | `hub/test/{soap,ssdp-advertiser,dlna-objects}.test.ts`                                                                  | `hub/test/sonos-control.test.ts` (imports moved to shared modules)                |
| Docs     | —                                                                                                                       | `docs/hub-internals.md`, `docs/system-architecture.md`, `example.env`             |

## Test plan

- [x] `pnpm --filter hub typecheck` clean
- [x] `pnpm --filter hub test` — 435 tests passing (30 new)
- [ ] Manual e2e on a host-networked deploy (requires Linux host + UDP multicast):
  - [x] `DLNA_ENABLED=true POUTINE_LAN_URL=http://<lan>:3000` boot logs `DLNA MediaServer enabled` and `DLNA SSDP advertiser started`
  - [ ] `curl http://<lan>:3000/dlna/device.xml` returns valid device description
  - [x] VLC → Local Network → "Poutine" appears, browsing reaches tracks, stream plays
  - [ ] Windows Media Player Legacy: server appears in `Other Libraries`, Artists/Albums browseable, MP3 tracks play
  - [ ] BubbleUPnP / Kodi smoke test — FLAC plays end-to-end
  - [ ] `ssdp:byebye` is observed on `tcpdump -i any -A udp port 1900` when the hub stops

## Out of scope (this PR)

- GENA event subscriptions — clients tolerate polling. No `/dlna/event/*` handlers yet.
- Transcoding to MP3/WMA for WMP Legacy compatibility. Tracked separately.
- DLNA+ profile beyond `MediaServer:1`.

## Base branch

This PR is based on `feature/108-sonos` (#162) because #175 explicitly stacks on the Sonos branch — the shared DIDL/SOAP/SSDP primitives originated there.


benders:
_FROM @claude:_

> Added LAN-only gate (commit e26af20).
>
> Every route under `/dlna/*` now goes through a `requireLan` preHandler (`hub/src/auth/lan-only.ts`) that returns 403 if the request carries any of these proxy-forwarding headers:
>
> - `x-forwarded-for`, `x-forwarded-host`, `x-forwarded-proto`, `x-real-ip`
> - `forwarded` (RFC 7239)
> - `cf-connecting-ip`, `cf-ray`
>
> Rationale: a public tunnel (Cloudflare Tunnel, Caddy, nginx, Tailscale Funnel, …) typically terminates on the same host as the hub and forwards traffic to loopback — from the hub's view those requests look identical to a LAN client. Real LAN clients (Sonos, WMP, Kodi, VLC) never set those headers, while every reasonable forward proxy sets at least one.
>
> Scope: `/dlna/*` only. `/cast/stream` keeps its HMAC token; Subsonic keeps its u+p / u+t+s auth — both already require authentication so they're safe to reach via the tunnel.
>
> 13 new tests in `hub/test/lan-only.test.ts`; full suite now 448 passing.

benders:
_FROM @claude:_

> Added integration tests (commit 7ffe31f). 19 integration tests pass via \`pnpm test:integration\`; full unit suite (448) still green.
>
> **\`test/dlna-ssdp.integration.test.ts\`** — boots SsdpAdvertiser, drives it from raw UDP in the test process.
> - M-SEARCH for MediaServer:1 → asserts USN / LOCATION / CACHE-CONTROL / SERVER on the unicast reply
> - M-SEARCH for ssdp:all → asserts all five advertised targets reply (rootdevice / uuid / MediaServer:1 / ContentDirectory:1 / ConnectionManager:1)
>
> **\`test/dlna-http.integration.test.ts\`** — boots the full hub via \`buildApp()\` on a random loopback port; \`poutineLanUrl\` left unset to keep the SSDP advertiser off (so it doesn't fight the SSDP test). Uses \`node-upnp\` as a real UPnP control point:
> - \`getDeviceDescription()\` → asserts MediaServer:1 device type, friendly name, UDN, and both required services present
> - \`call('ContentDirectory', 'Browse', {ObjectID: '0', ...})\` → asserts NumberReturned / TotalMatches and decoded DIDL-Lite shape
> - \`call('ConnectionManager', 'GetProtocolInfo')\` → asserts http-get audio profiles
> - Raw fetch for single-output SOAP actions (GetSortCapabilities, GetSearchCapabilities, GetSystemUpdateID), \`/dlna/stream/<unknown>\` 404, and the LAN-gate rejecting x-forwarded-for / cf-connecting-ip
>
> **Library notes:**
> - **\`node-upnp\`** (added as devDependency) — kept. Pays off for device description + Browse; the SOAP request/response flow reads as a real UPnP call. One quirk: \`parseSOAPResponse\` walks the SCPD argumentList via \`Array.from(argumentList.argument)\`, which silently returns \`[]\` for single-argument actions because fast-xml-parser emits an object (not an array) for a single child. Raw fetch covers those.
> - **\`@achingbrain/ssdp\`** — evaluated and dropped. Its \`discover()\` async iterator only yields after fetching+parsing the LOCATION through an internal pipeline, which adds failure modes orthogonal to what we want to test. Raw \`dgram\` gives an unambiguous packet-level assertion and avoids the UDP-1900 reuse-port conflict you get when both the bus and the advertiser bind to 1900.
>
> CI note: both integration test files require UDP multicast (SSDP) and a free UDP 1900 (advertiser) and HTTP listen on loopback — fine on macOS / Linux dev hosts and Linux CI runners with host networking; will not work in sandboxed networking setups (matches the existing \`*.integration.test.ts\` exclusion pattern).

benders:
FROM @claude:

> ## Review — code quality + risk pass
>
> Scaffold is well-organized; helpers extracted to `services/{soap,didl}.ts` cleanly. SSDP, SOAP, object-ID parsing have solid unit coverage and the integration test drives a real UPnP client. The LAN gate is a nice safety net. Main concerns below — work to follow tomorrow.
>
> ### Headline: stream handler duplication
> `routes/dlna.ts:196-353` reimplements ~150 lines of `routes/cast.ts` — source selection, `SubsonicClient` vs `federatedFetch` branching, `stream-tracking` start/update/finish, response-header forwarding, byte counting. Only deltas: attribution user (`dlnaPseudoUser` vs token-recovered username), the `kind` tag, and two extra DLNA response headers. **Recommendation:** factor into `services/stream-relay.ts` exposing something like `relayTrackStream({ db, app, trackId, username, kind, queryParams, requestHeaders, extraResponseHeaders }) -> reply pipeline`. Both `cast.ts` and `dlna.ts` shrink to ~30 lines each.
>
> ### `services/ssdp-advertiser.ts`
> - `start()` calls `socket.bind(SSDP_PORT)` and `addMembership(SSDP_ADDR)` with no interface argument. On a multi-NIC host the kernel picks one — likely the wrong one for a Docker host with both `docker0` and the LAN NIC. Consider explicit `addMembership(SSDP_ADDR, interfaceAddress)`, derived from `POUTINE_LAN_URL`'s host.
> - `stop()` → `announceByebye()` fires UDP packets then immediately calls `socket.close()` synchronously. On Linux the unsent packets in the socket buffer may be discarded, defeating the byebye. Either await `send` callbacks before `close()`, or send byebye 2–3× (common SSDP idiom) so at least one reaches the wire.
> - `parseMSearch`: `parseInt(headers.mx || \"0\", 10)` then `parsed.mx * 1000` — if MX is missing the delay collapses to 0 which actually matches spec (immediate), but a malformed MX like `\"abc\"` yields NaN → `Math.min(1000, NaN) = NaN` → `Math.random()*NaN = NaN` → `setTimeout(fn, NaN)` fires immediately. Behaviorally fine, but a `Number.isFinite` guard would be defensive.
> - `TARGETS(uuid)` rebuilds on every M-SEARCH + every alive tick. Cache once in the constructor.
>
> ### `routes/dlna.ts`
> - `soapFault` uses HTTP 500 with `upnpCode 401` (\"Invalid Action\"). Per UPnP-DA §3.2.2 that's correct — but worth a one-line comment in `soapFault` so readers don't misread it as an HTTP-auth code.
> - `pickXmlTag` on the SOAP body is regex-based, non-recursive. Fine for CDS:1 Browse args (all leaf strings) but will silently mis-pick if a future action embeds nested same-named tags. Document the limitation in `soap.ts`, or add a real XML parser when ContentDirectory `Search` is implemented.
> - No `bodyLimit` set on this plugin scope. The default Fastify limit (1 MB) is fine but explicit `app.register(dlnaRoutes, { bodyLimit: 64 * 1024 })` makes the SOAP-only nature obvious and limits abuse.
> - `UpdateID: \"1\"` is hardcoded for both `Browse` and `GetSystemUpdateID`. DLNA clients (notably WMP and Kodi) use this to invalidate their object cache after library sync. v1 is OK, but add a TODO referencing a bump on `auto-sync` completion — otherwise a renamed album never re-displays without a server re-add.
> - `app.config.dlnaPseudoUser || app.config.poutineOwnerUsername || \"dlna\"` — if neither is set and a peer stream is requested, `federatedFetch({ asUser: \"dlna\" })` may not resolve at the peer side. Either fail-fast at boot when `DLNA_ENABLED=true && !poutineOwnerUsername && !dlnaPseudoUser`, or document the constraint.
>
> ### `auth/lan-only.ts`
> - `requireLan` sends the 403 but doesn't `return reply` — relies on Fastify halting subsequent handlers when a reply is sent in a hook. Explicit `return reply` (or `await reply`) reads more obviously and matches the project's other preHandlers (`requireAuth`).
> - The blocklist is good, but worth noting `via` (RFC 7230, set by some forward proxies) and `x-cluster-client-ip` aren't covered. Not a blocker; the listed seven catch every common case. Consider an allow-listed env override (`DLNA_TRUSTED_PROXY_HEADERS=`) for operators who run a LAN proxy.
>
> ### Tests
> - Strong unit coverage on helpers; `dlna-objects.test.ts` seeds the real schema (good).
> - **Gap:** `/dlna/stream/:trackId` has only the 404 path tested. Add a positive case (local source + Range header) to lock in the response-header contract — especially `contentFeatures.dlna.org`. The existing `hub/test/stream.test.ts` fake-Navidrome harness should be reusable once the relay helper is extracted (see headline).
> - **Gap:** `ssdp-advertiser` integration test asserts M-SEARCH response but not the `ssdp:byebye` on `stop()`. A `dgram` socket listening to `239.255.255.250:1900` during teardown would close the loop on the byebye fix above.
> - `dlna-http.integration.test.ts:54-56` deliberately leaves `poutineLanUrl` unset to avoid UDP-1900 contention with the other integration test. Document this constraint in `docs/dlna.md` testing section so the next person doesn't re-flake it.
>
> ### Smaller / polish
> - `xmlEscape`'s ternary chain (`soap.ts:8-15`) is hard to read. A 5-entry object literal lookup is shorter and obviously a 1:1 map.
> - `dlna-objects.ts:64`: `${ARTISTS_ID.replace(\"/artists\", \"/artist\")}/${id}` is cute but `\`0/music/artist/\${id}\`` is clearer and matches the parse regex above it.
> - `routes/dlna.ts:225-227`: `if (!best || !trackRow)` returns 404 — but a missing `track_sources` row for an existing track means \"no preferred source available,\" which is a 503/409, not 404. Worth distinguishing for operator triage.
>
> Will tackle these tomorrow. Headline (relay extraction) + byebye fix are the two I'd prioritize before flipping out of Draft.

benders:
FROM @claude:

> ## Review concerns addressed
>
> Three commits on top of the existing branch — all 471 unit tests still passing.
>
> **a747dcd — extract shared stream relay** (headline). \`services/stream-relay.ts\` owns source-selection, transcoding decision, \`stream-tracking\` bookkeeping, and the upstream-to-reply pipe. \`cast.ts\` shrinks to ~40 lines (token auth → relay). \`dlna.ts\` stream handler shrinks to ~25 lines (resolve pseudo-user → relay with extra DLNA response headers). Distinguishes 503 (no preferred source) from 404 (unknown track) for both endpoints. Adds \`dlna-stream.test.ts\` — fake-Navidrome positive path locking in \`contentFeatures.dlna.org\` / \`transferMode.dlna.org\` headers, Range forwarding, and the 404/503 split.
>
> **1d6cb36 — SSDP advertiser + LAN-gate clarity.**
> - \`addMembership\` now pins an explicit IPv4 interface (derived from \`POUTINE_LAN_URL\`'s host with a non-internal fallback) — fixes the multi-NIC docker0/LAN-NIC race.
> - \`stop()\` is async, sends byebye 3× per target and awaits each send-callback before \`socket.close()\` — packets actually leave the wire now. \`server.ts\` \`onClose\` awaits.
> - \`parseMSearch\` coerces non-numeric / negative MX to 0 — no more NaN into \`setTimeout\`.
> - \`TARGETS\` cached in the constructor.
> - \`requireLan\` returns the reply explicitly.
> - New byebye integration test (boots a second advertiser, joins the multicast group, asserts all 5 byebye NTs arrive before close). New unit cases for the MX coercion.
>
> **6adf491 — fast-xml-parser + small polish.**
> - \`pickXmlTag\` now uses \`fast-xml-parser\` instead of the regex — safe for future \`Search\` action's nested / CDATA-bearing args. The 11 \`soap.test.ts\` cases still pass unchanged.
> - \`xmlEscape\` rewritten as object lookup.
> - \`artistObjectId\` / \`albumObjectId\` use literal path strings.
> - \`docs/dlna.md\` picks up the new test row + a note documenting the \`POUTINE_LAN_URL\` unset constraint in \`dlna-http.integration.test.ts\` (UDP-1900 contention).
>
> SOAP control endpoints now have a 64 KB \`bodyLimit\` (per-route, since Fastify v5 doesn't accept \`bodyLimit\` in \`register()\` options). \`soapFault\` carries a header comment so a future reader doesn't misread the inner \`upnpCode: 401\` as HTTP Unauthorized.
>
> ## Follow-ups filed
>
> - #187 — bump \`SystemUpdateID\` after \`auto-sync\` runs (current \`\"1\"\` is a TODO).
> - #188 — fail-fast at boot when \`DLNA_ENABLED=true\` and neither \`DLNA_PSEUDO_USER\` nor \`POUTINE_OWNER_USERNAME\` is set (peer streams break silently).
> - #189 — optional \`DLNA_TRUSTED_PROXY_HEADERS\` allow-list for transparent LAN-side reverse-proxy deployments.
>
> Ready for re-review. Still in Draft pending your call on flipping to Ready.

PR #185, opened 2026-05-16
Reflect local Navidrome reachability in /api/health

### LINKED ISSUE #178: Hub /api/health should reflect local Navidrome reachability
## Problem

`GET /api/health` currently returns `status:"ok"` regardless of whether the hub can reach the local Navidrome. This means the endpoint reports healthy when the proxy is effectively broken — every `/proxy/rest/*` call will 500 with `ECONNREFUSED`, but health monitors / load balancers / the federation handshake all see green.

Surfaced while validating PR #162 (Sonos host networking override): hub couldn't reach Navidrome at the configured URL, but `/api/health` happily returned 200.

## Proposed shape

```json
{
  "status": "ok" | "degraded",
  "appVersion": "...",
  "apiVersion": 5,
  "navidrome": "ok" | "unreachable"
}
```

- Probe Navidrome via a cheap endpoint (e.g. `GET /ping`) with a short timeout (~1s) on each health check.
- If unreachable, set `status:"degraded"` and `navidrome:"unreachable"`.
- Always return HTTP 200 so the federation handshake (`/federation/handshake` reads `apiVersion` / `appVersion` from the peer's `/api/health`) continues to work when a peer's Navidrome is briefly down. Federation only validates apiVersion.

## Notes

- Don't cache the probe result — health checks are expected to reflect current state.
- Don't expose Navidrome error details (URL, error message) — keep the response body to the two-value enum.

benders:
## Summary

- `GET /api/health` now probes the local Navidrome via Subsonic `/rest/ping` (~1s timeout) on each request and reflects reachability in the response body: `{ status: "ok" | "degraded", appVersion, apiVersion, navidrome: "ok" | "unreachable" }`.
- Always returns HTTP 200 so the federation handshake (which reads `apiVersion` / `appVersion` from a peer's `/api/health`) keeps working when a peer's Navidrome is briefly down — peers only validate `apiVersion`. Operators / load balancers should key on `body.status`.
- Updates `docs/hub-internals.md` route table to document the new shape.

closes #178

## Test plan

- [x] New `hub/test/health-route.test.ts` covers both branches (unreachable → `status:"degraded"`, `navidrome:"unreachable"`; reachable mock Subsonic ping → `status:"ok"`, `navidrome:"ok"`).
- [x] Updated `static-serving.test.ts` assertions for the new body shape.
- [x] `pnpm verify` — hub 423/423 + frontend 62/62 passing.
- [ ] Manual: hit `/api/health` against a running hub with Navidrome up, then stop Navidrome and re-check — confirm `status` flips to `"degraded"`.


benders:
FROM @claude: Review of #185

> **Overview**
> Adds a Subsonic `/rest/ping` probe to `/api/health` with a 1s timeout, surfaces it as `navidrome: ok|unreachable` and `status: ok|degraded`. Always 200 so federation handshake survives a peer's Navidrome blip. Two new tests cover both branches; existing static-serving tests updated for new shape. Docs row updated.

> **Correctness / risks**
>
> - **Dangling fetch on timeout.** `Promise.race` with a `setTimeout` rejecter resolves the route, but the underlying `fetch` inside `SubsonicClient.ping()` keeps running to completion in the background (and its eventual rejection becomes an unhandled rejection if Navidrome later errors). Prefer `AbortSignal.timeout(1000)` plumbed through `SubsonicClient.request` (add an optional `signal`), or wrap in `AbortController` and `controller.abort()` in the `finally`. This also cleans up the `clearTimeout` dance.
> - **No request coalescing / caching.** Every `/api/health` hit fires a fresh Subsonic ping. LB/monitoring probes can hammer this (and through it, Navidrome). Consider a small in-process cache (e.g. memoize result for 1–2s) — keeps probe semantics for humans while bounding amplification.
> - **New `SubsonicClient` per request.** Cheap, but the hub already wires one up elsewhere; reusing the existing instance (or hanging one off `app`) keeps construction out of the hot path and matches convention.
> - **`String(err)` drops the stack.** `app.log.warn({ err }, ...)` lets pino serialize it properly.
>
> **Tests**
>
> - `127.0.0.1:1` for "refused" is usually fast on Linux/macOS but not contractually guaranteed (some sandboxes route differently). A `net.createServer().listen(0)` immediately `.close()`d, or pointing at a server that `socket.destroy()`s on connect, is more portable. Low priority — flag if CI ever flakes here.
> - Worth one test that asserts the 1s timeout actually fires (fake server that `setTimeout`s the response past the budget) — currently the timeout path is only exercised via connection refusal, not a slow peer, which is the more realistic failure mode this PR exists to catch.
>
> **Docs / contract**
>
> - Route-table cell is dense; the rationale ("always 200 so federation handshake…") reads better as a short note under the table than inline. Not blocking.
> - Worth a one-liner in `docs/pitfalls.md`: "`/api/health` 200 ≠ healthy — key on `body.status`." Easy trap for future operators wiring k8s/LB probes.
>
> **Style nit**
> Long block comment above the handler duplicates the PR description; the `issue #178` reference + a one-line "always 200, see docs" is plenty — body of the reasoning belongs in the docs row (already there).

> Non-blocking overall. The abort-vs-race fix is the only one I'd want before merge.

Issue #224, opened 2026-05-26
Sonos cast fails when POUTINE_OWNER creds don't match a hub user


benders:
## Problem

Sonos `/api/sonos/devices/:id/play` returns 404 \"Track not found\" when \`POUTINE_OWNER_USERNAME\` / \`POUTINE_OWNER_PASSWORD\` in \`.env\` don't correspond to a real hub Subsonic-authable user.

### Trace
1. SPA → \`POST /api/sonos/devices/:id/play { trackId: \"t<uuid>\" }\`
2. \`buildCast\` calls \`hubSubsonicSonos.call(\"/rest/getSong\", { id })\`
3. \`HubSubsonicCaller\` auths via \`POUTINE_OWNER_USERNAME\` + \`POUTINE_OWNER_PASSWORD\` (plaintext \`u+p\`)
4. Hub Subsonic auth rejects creds → \`subsonic-response.status=failed, error.code=40\`
5. \`buildCast\` interprets failed status as \"track not found\" → 404
6. SPA shows cast failure

### Observed
- Native macOS hub instance with admin user \`xac\`, env \`POUTINE_OWNER_USERNAME=admin\` (Navidrome admin, not a hub user) → every Sonos cast 404s.
- Same code on Docker instance where admin login user *is* \`admin\` works fine.

## Fix (recommendation B from design discussion)

Eliminate the owner-cred path for Sonos. Sonos routes already run under \`requireAuth\` — \`req.username\` is known. Plumb the caller's identity through to \`HubSubsonicCaller\` and add a trusted in-process auth mode:

- \`HubSubsonicCaller.call(endpoint, params, opts?)\` gains optional \`{ asUser: string }\`
- When \`asUser\` set: inject with headers \`x-poutine-internal: <secret>\` + \`x-poutine-as-user: <username>\` instead of \`u+p\`
- Hub Subsonic auth middleware recognizes the shared-secret header and resolves the user without password check
- \`routes/sonos.ts\` passes \`req.username\` to \`buildCast\` → caller
- DLNA browse still uses owner-cred mode (no user context — LAN-gated, unauthenticated)

## Acceptance
- Sonos cast succeeds even when \`POUTINE_OWNER_*\` is misconfigured
- \`HubSubsonicCaller\` supports both per-user (trusted-header) and owner (u+p) modes
- New auth path documented in \`docs/authentication.md\`
- Pitfall added to \`docs/pitfalls.md\`
[...]

Issue #221, opened 2026-05-21
Phase 7: enforce Hub/Player directory boundary in CI


benders:
> FROM @claude:
>
> Parent: #212
> Depends on: #219
>
> ## Goal
>
> Make the boundary mechanically enforced. Future PRs can't accidentally reintroduce cross-context imports.
>
> ## Scope
>
> - Reorganize \`hub/src/\` (or repo root) into directories that match the bounded contexts: \`hub/\`, \`player/\`, \`shared/\` (and optionally \`contracts/\`).
> - Add \`eslint-plugin-import\` with \`no-restricted-paths\` (or \`dependency-cruiser\` if richer rules wanted):
>   - \`player/**\` may not import from \`hub/**\`.
>   - \`hub/**\` may not import from \`player/**\`.
>   - Both may import from \`shared/**\` and \`contracts/**\`.
> - CI fails on violation.
> - Optionally promote to TypeScript project references (separate \`tsconfig.json\` per context) for compile-time enforcement.
>
> ## Done when
>
> - Lint rule is in place and passing on \`main\`.
> - A deliberate test violation (in a throwaway commit) is caught by CI.
> - Doc note in \`docs/system-architecture.md\` describes the boundary.

benders:
FROM @claude:

> Phase 7 landed on `feature/212-rearchitect-hub-player-boundary` (merge commit on the parent branch after `phase/221-ci-boundary`).
>
> ## What shipped
>
> **Backend (`hub/eslint.config.js`)** — Player-side files may not runtime-import:
> - `better-sqlite3` (type-only via `allowTypeImports: true` is fine)
> - `../adapters/subsonic` (the in-process `SubsonicClient`)
> - `**/db/preferred-source*`, `**/db/client*`, `**/db/schema*`
>
> Player-file glob: `routes/sonos.ts`, `routes/dlna.ts`, `services/sonos-*.ts`, `services/dlna-*.ts`, `services/cast-tokens.ts`, `services/didl.ts`, `services/soap.ts`, `services/ssdp-advertiser.ts`, `services/player-settings.ts`.
>
> **Carve-out:** `hub/src/db/player-db.ts` (the player.db opener) is excluded from the `playerFiles` rule scope — it must import `better-sqlite3` to open the DB.
>
> **Frontend (`frontend/eslint.config.js`)** — no cross-imports between `features/hub-admin/`, `features/player-admin/`, and `features/player/`. `features/shared/` is unrestricted on purpose.
>
[...]

Issue #220, opened 2026-05-21
Phase 6: remove in-process Subsonic adapter use from Player routes


benders:
> FROM @claude:
>
> Parent: #212
> Depends on: #218
>
> ## Goal
>
> Player code stops importing \`SubsonicClient\` from \`hub/src/adapters/subsonic.js\`. All Player → Hub Subsonic calls go over HTTP, not in-process function calls. This is the final boundary cleanup before lint enforcement.
>
> ## Scope
>
> - \`routes/sonos.ts\` switches its remaining \`SubsonicClient\` imports to the HTTP client introduced in Phase 5.
> - Audit any other Player-side code for in-process Hub adapter use; convert all to HTTP.
> - Shared auth helper (JWT verify) moves to \`shared/\` if not already — Player must not import from \`hub/src/auth/\` except via the shared verifier.
>
> ## Done when
>
> - \`grep -r "from .*hub/src" player code\` (or the future \`player/\` directory) returns only \`shared/\` imports.
> - All Player → Hub traffic visible as HTTP calls in logs.

benders:
FROM @claude:

> Starting Phase 6. Worktree at `/private/tmp/poutine-220`, branch `phase/220-remove-inproc-subsonic`.
>
> Plan:
>
> 1. Replace `SubsonicClient` import in `hub/src/routes/sonos.ts` with an HTTP loopback caller — same `app.inject()` pattern Phase 5 (#219) introduced for DLNA. The hi-res FLAC probe (only remaining in-process Subsonic use in Player code) routes through Navidrome via the loopback shape.
> 2. Add `/api/admin/hub/*` and `/api/admin/player/*` namespace aliases. Existing `/admin/*` paths stay live; the new namespaces mount the same handlers at the per-ownership prefix per #212. Frontend `lib/api.ts` switches admin calls to the new paths.
> 3. Update docs (`system-architecture.md`, `sonos.md`) to record the HTTP-only Sonos path.
>
> Deferring to a follow-up issue (against #212):
> - Remove `getPreferredSource` + raw `app.db` reads from `routes/sonos.ts`. Doing so requires Hub `/rest/getSong` to expose `samplingRate`/`bitDepth` and the underlying schema column to land (called out by #199 in code). Separate from the SubsonicClient cut.

benders:
FROM @claude:

> Starting phase 6 work. Surveyed state on parent feature branch (557ba5d after #219 merge):
>
> Player-code violations remaining:
[...]

Issue #219, opened 2026-05-21
Phase 5: rewrite DLNA ContentDirectory against Subsonic HTTP client


benders:
> FROM @claude:
>
> Parent: #212
> Depends on: #213 (spike), #214 (OpenSubsonic survey)
>
> ## Goal
>
> Replace \`hub/src/services/dlna-objects.ts\` (540 lines, direct \`better-sqlite3\` access to library tables) with an implementation that reads only via Subsonic HTTP calls. After this phase, Player BE has zero \`better-sqlite3\` imports.
>
> ## Scope
>
> - Stand up a Subsonic HTTP client in Player (\`player/subsonic-client.ts\`), authenticated with the service token from \`player.db\`.
> - Rebuild \`dlna-objects.ts\` to query the client instead of SQL. Map browse/search/metadata operations 1:1 to Subsonic calls per the spike (#213) findings.
> - Apply any OpenSubsonic extensions recommended in #214.
> - Use in-process HTTP (loopback or \`app.inject()\`) while combined — same code path for future deploy split.
> - Cache where measurements (from #213) show it's necessary; document each cache with a TTL rationale.
>
> ## Done when
>
> - \`grep -r better-sqlite3 hub/src/{services,routes}/{sonos,dlna,cast}*.ts hub/src/services/stream-relay.ts\` returns no matches.
> - DLNA browse + search behave identically (per spike acceptance criteria).
> - Live instance tested with at least one real DLNA renderer.

benders:
FROM @claude:

> ## Phase 5 complete — DLNA on Subsonic, no DB
>
> Branch `phase/219-dlna-subsonic` merged into `feature/212-rearchitect-hub-player-boundary` (557ba5d).
>
> ### Implementation
>
> - `hub/src/services/dlna-objects.ts` rewritten on the `SubsonicCaller` interface. No `better-sqlite3` import anywhere in DLNA, Sonos, or cast modules — the done-criteria grep is empty.
> - Production `SubsonicCaller` lives in `server.ts`: `app.inject()` over loopback `/rest/*`, owner u+p, `c=poutine-dlna`. HTTP-shaped so a future split deploy swaps it for a real loopback fetch.
> - `Browse` and `Search` are now async; `routes/dlna.ts` awaits.
> - Spike module (`dlna-objects-subsonic.ts`) and spike test (`dlna-subsonic-spike.test.ts`) deleted; the production module + production test took their place.
>
> ### "All Tracks" decision
>
[...]

PR #225, opened 2026-05-26
Hub/Player boundary re-architecture (#212)

benders:
## Summary

Completes #212: enforces a hard Hub/Player split inside the same process.

- **#213-#214** — DLNA-over-Subsonic spike + OpenSubsonic extensions survey.
- **#215** — `player.db` introduced (DLNA UUID, cast signing key, Player settings live here, not in hub.db).
- **#216** — Admin SPA split into Hub and Player destinations.
- **#217** — Runtime Player settings migrated into `player.db`.
- **#218** — Stream relay eliminated; cast/DLNA devices receive Hub Subsonic URLs directly (cast tokens promoted to first-class auth on `/rest/stream`).
- **#219** — DLNA `ContentDirectory` rewritten on Subsonic HTTP; no DB access from Player.
- **#220** — Sonos routes off the in-process Subsonic adapter; `/api/admin/{hub,player}/*` namespaces.
- **#221** — Lint + CI enforcement of the Hub/Player directory boundary (backend + frontend).
- **#224** — Trusted in-process Subsonic auth path (`x-poutine-internal` + `x-poutine-as-user`) so Player code authenticates as the calling SPA user instead of POUTINE_OWNER. Removes the silent-404 Sonos cast regression.

Player-side code (sonos, dlna, player-admin) no longer touches `app.db` or imports `SubsonicClient` — everything flows through `HubSubsonicCaller` over `app.inject`. Boundary is verified by `lint:boundary` in CI.

## Test plan

- [x] \`pnpm verify\` — 595 hub tests + 96 frontend tests
- [x] \`pnpm lint:boundary\` — backend + frontend boundary checks
- [x] Live Sonos cast against real device (hi-res regression tracked separately under #199)
- [ ] DLNA browse + stream from a real renderer
- [x] Verify SPA admin pages load under both \`/admin/hub/*\` and \`/admin/player/*\`


benders:
> FROM @claude:
>
> Two follow-up commits pushed on top of the existing PR before review:
>
> - **ba65aeb** — frontend lint cleanup. Three `react-hooks` / `react-refresh` errors that pre-dated #212 are now resolved (DevicePicker render-phase pattern, AlbumsPage prev-value reset, ErrorMessage formatError extraction). Two surviving `exhaustive-deps` warnings in PlayerBar are intentional false-positives (re-keying on the full `currentTrack` re-triggers #185) and now carry inline rationales. `AGENTS.md` per-task checklist updated to require zero lint output as a gate, alongside `lint:boundary`.
> - **46eede5** — docs audit pass against the post-#212 architecture. Trimmed duplication so each fact lives in one place: `system-architecture.md` owns the dual-DB layout and admin-mount table, `sonos.md` / `dlna.md` own their respective surfaces, `hub-internals.md` links out instead of restating. Fixed a stale claim in `sonos.md` that still pointed `sonos_enabled` / `sonos_volume_cap` at `hub.db.settings` (true pre-#217, wrong now). `frontend-testing.md` no longer describes the runtime boundary walker as a placeholder for ESLint — #221 has landed.
>
> Filed **#226** for the admin-mount handler-partition cleanup — the only follow-up from the original wrap-up comment that wasn't already covered (#199 had merged earlier).
>
> `pnpm verify` + `pnpm lint` both clean.

benders:
FROM @claude: research notes on review findings 1 and 4.

**Finding 1 — `requireSubsonicAuthBinary` and internal auth:** verified wired.

> `hub/src/auth/subsonic-auth.ts:194-202` calls `tryInternalAuth(request)` after the cast-token branch and before falling through to `u+p`/`u+t+s`. Same discriminated `ok`/`fail`/`absent` handling as `requireSubsonicAuth`, just with `sendBinaryError(reply, 401, ...)` instead of the Subsonic envelope. No gap — `HubSubsonicCaller` can hit binary endpoints `asUser` safely.

**Finding 4 — `asUser` user lookup:** verified safe; one minor note.

> `tryInternalAuth` (subsonic-auth.ts:105-117) looks the user up by `username` against `users`, returns `{ kind: "fail" }` when no row exists, and sources `isAdmin` from `user.is_admin === 1` — i.e. straight from the DB, never the request. `username` is `UNIQUE NOT NULL` per `hub/src/db/schema.sql`, so no ambiguity.

> No disabled/locked column exists in the `users` schema today, so there is nothing to enforce on that axis. If a future change adds account-disabled / soft-delete state, this lookup needs the same `WHERE ... AND disabled = 0` clause as `lookupAndVerify` will (neither path enforces it today, so they stay symmetric).

Both items cleared. The two asks from the review are resolved; remaining items in the review are nice-to-have hardening, not blockers.

Issue #1, opened 2026-04-11
Move tasks to GH Issues


benders:
Currently using local Yaks task tracking system. Migrate to GitHub Issues for better visibility and collaboration.

benders:
[@claude] Migration complete.

- Created labels: p1, p2, p3, idea, chore
- Migrated 12 open (hairy) yaks to GH Issues: #2–#13
- Skipped 8 shorn yaks (already done)
- Updated AGENTS.md: Yaks → GH Issues, added `[@claude]` prefix rule for agent comments
- Removed `.yaks/` directory

Awaiting user approval to commit.

benders:
[@claude] Done. Committed 11855db.

PR #235, opened 2026-05-29
feat(play-counts): track plays across local + peer media, expose via Subsonic (#197)

### LINKED ISSUE #197: Track play counts for local + peer media, expose via Subsonic API
Record per-user play events for every track played through Poutine — regardless of whether the source is the local Navidrome or a peer hub — and surface the data through the Subsonic API so clients can show "most played", "play count", and "last played".

## Motivation
Navidrome already tracks plays for its own library, but those counts are siloed per-hub and miss anything streamed from peers. Poutine is the unifying layer, so it should own the canonical play history across the federated catalog.

## Scope
- New table (e.g. `play_events`): user_id, unified_track_id, source_instance_id, played_at, duration_played (for scrobble threshold)
- Aggregate views: per-track count, per-user count, last-played timestamp
- Recording trigger: SPA reports playback milestones (e.g. > 50% or > 4 min, mirroring Last.fm rules) to a hub endpoint
- Cast targets (Sonos/DLNA) also report — capture from server-side stream proxy when SPA isn't the playback surface
- Subsonic API: populate `playCount` and `played` fields on `getSong`, `getAlbum`, `getAlbumList` (type=frequent / recent), `getRandomSongs` filters, etc.
- Federation: optional — should peer hubs share aggregate counts? (out of scope for v1; track separately if pursued)

## Open questions
- Per-user vs. per-instance counts in Subsonic responses (Subsonic spec is per-user)
- Backfill: import existing Navidrome play history on first run?
- Retention policy for raw event log vs. rolled-up aggregates

benders:
Closes #197.

Adds a durable, per-user play history spanning the merged catalog (local + peer media) and surfaces it through the Subsonic API. Poutine — not the siloed per-hub Navidrome counts — becomes the canonical owner of play history across the federation.

## What's in v1 (scope confirmed with owner)
- **New `play_events` table** — per-user, source-aware, durable (never pruned). Distinct from the ephemeral `stream_operations` activity feed. No FK on `unified_track_id`; orphans dropped at read time like `user_stars`. Millisecond `played_at` so `recent` ordering is stable within a second.
- **`/rest/scrobble` implemented** (was a no-op stub). `submission=true` (default) records; `submission=false` is now-playing only. Batch-tolerant: unknown/malformed ids skipped. **This is the single play-recording path** — the server persists what the client reports and never infers plays from stream lifetime.
- **SPA reporting** — `PlayerBar` fires one scrobble per play when playback crosses a Last.fm-style threshold (≥50% or >4 min), for **both** local `<audio>` and Sonos-cast playback. Sonos uses the device's real reported transport position (`getSonosState` → `s.position`), so casts count at the same threshold with no server-side estimate. One `scrobbledRef` guard covers both paths so each play counts once.
- **Exposure** — per-user `playCount` + `played` projected onto `getSong`, `getAlbum`, `getAlbumList2`, `getArtist`, `search3`, and the starred endpoints. `getAlbumList2?type=frequent`/`recent` rank by play history and exclude never-played albums.

## DLNA
Not wired for play-counts (not currently in use). When DLNA is reinstated it gets its own position-reporting path; it must not resurrect a server-side wall-clock estimate. `StreamTrackingService` records no plays at all.

## Out of scope (tracked separately)
- Navidrome play-history backfill on first run (counts start at zero from deploy).
- Federating aggregate counts between peers.

## Testing
- `hub/test/play-events.test.ts` — `PlayEventService` unit + scrobble/projection integration (per-user scoping, frequent/recent ordering, batch resilience).
- `hub/test/stream-tracking.test.ts` — activity-feed behavior (no longer records plays).
- `frontend/.../PlayerBar.test.tsx` — Sonos scrobble fires past threshold, stays silent below it.
- `pnpm verify` (typecheck + lint:boundary + 665 hub / 99 frontend tests) and `pnpm lint` both green.

## Docs
`docs/opensubsonic.md`, `docs/hub-internals.md` (new "Play counts" section), `docs/pitfalls.md`, `docs/system-architecture.md`.


benders:
FROM @claude:

> **Update — Sonos casts now scrobble from the SPA; server-side estimate removed.**
>
> Per review direction, collapsed to a single play-recording path. The SPA already polls the Sonos device's real transport position (`getSonosState` → `s.position`) while casting, so it now scrobbles at the same Last.fm threshold as local `<audio>` playback — accurate playback position rather than the stream connection lifetime.
>
> - `PlayerBar` fires `/rest/scrobble` for both local and Sonos-cast playback, guarded by one `scrobbledRef` so each play counts once.
> - `StreamTrackingService` no longer records plays: dropped the `PlayEventService` injection, the `userId` stream-op field, and `PlayEventService.recordIfThreshold` (the threshold decision lives entirely in the client now).
> - DLNA is not currently in use; it'll get its own position-reporting path when reinstated and must not bring back a server-side wall-clock estimate.
>
> The "known cast/DLNA approximation" caveat from the original PR description **no longer applies** — Sonos counts are now exact.
>
> Verified: `pnpm verify` (665 hub + 99 frontend tests, typecheck, lint:boundary) and `pnpm lint` all green. Commit `0f4d7ca`.

benders:
FROM @claude:

> ## Code review (`/code-review` medium effort) — 8 findings
>
> Verified all candidates against the code (Phase-2 verifier pass). Correctness ranked ahead of cleanup. The first three cluster on the client-side scrobble logic.
>
> ### Correctness
>
> **1. Sonos poll scrobbles with no `PLAYING` gate → phantom plays** — `PlayerBar.tsx:464`
> The poll fires `scrobble()` on `pos >= threshold` alone. A paused/seeked cast keeps its reported position, so the next 1.5s tick records a play that was never heard. The local `<audio>` path can't do this (timeupdate only fires during playback), so the two paths disagree.
>
> **2. Unconditional `scrobbledRef` reset → Sonos→local double-count** — `PlayerBar.tsx:175`
> The local stream-load effect resets `scrobbledRef` after the `if (isSonos) return` guard. A mid-track Sonos→local switch re-fires it, re-arms, and `handleTimeUpdate` re-scrobbles a play already counted on Sonos. The reverse direction is guarded by `if (resumeAt < 1)`; this direction isn't.
>
> **3. Resume near track-end scrobbles on first tick** — `PlayerBar.tsx:465`
> When `base > 0`, threshold is computed from full track length but compared to `s.position + base`, which already includes the offset. Resuming a cast at 180s of a 200s track scrobbles immediately though nothing was heard on this device.
>
> **4. No scrobble-on-end fallback → fully-played track can be lost** — `PlayerBar.tsx:611`
> `handleEnded` resets `scrobbledRef` and calls `next()` but never scrobbles; `handleTimeUpdate` is the only local trigger. A throttled/backgrounded tab can miss the threshold tick and reach `ended` uncounted.
>
> **5. Folder-scoped "Most Played" ranks by cross-source counts** — `subsonic.ts:753`
> The `playJoin` aggregates plays with only `WHERE pe.user_id = ?` — no instance filter — while the outer query scopes the album set by instance. A `musicFolderId` + `type=frequent` request (the SPA does this) ranks scoped albums by their global play counts.
>
> **6. Scrobble ignores Subsonic `time` param → backfill mis-stamped** — `subsonic.ts:1704`
> Handler never reads `time`; `PlayEventService.record` hard-codes `played_at` to `strftime('now')`. Offline/batched plays submitted later land at sync time, corrupting `type=recent` ordering and `played`.
>
> ### Maintainability / efficiency
>
> **7. Stale comment claims `finish()` records cast/DLNA** — `subsonic.ts:1701`
> Contradicts `stream-tracking.ts` (UPDATE only, no `PlayEventService`) and `docs/pitfalls.md` (one path only). Invites a future double-count or lost-count regression.
>
> **8. `getAlbumList2` frequent/recent double-aggregates full history** — `subsonic.ts:746`
> The `playJoin`'s `COUNT(*)`/`MAX(played_at)` are computed over the whole history for ORDER BY, discarded (never selected into the row), then re-computed by `annotatePlays`→`getAlbumStats` for the page. Select and reuse them.
>
> ### Cut for the 8-cap (noted, lower severity)
> - `frequent`/`recent` `ORDER BY` lacks a unique tiebreaker (`urg.id`) → tied albums can dup/drop across offset pages; gated behind >500 played albums.
> - Threshold math (`min(len/2, 240)` + guard-and-fire) is copy-pasted between the local and Sonos paths — extract one position+state-aware helper (also the root of #1/#3).
> - AlbumsPage re-enumerates the server-ordered set in three places.
>
> **Refuted:** `subsonicUser.id` deref (route is behind `requireSubsonicAuth`); multi-release album inflation (`unified_tracks.release_id` is singular).
>
> Starting on the scrobble cluster (#1–#4) now — a single position+state-aware scrobble helper resolves #1, #2, #3 and makes room for the #4 end fallback.

benders:
FROM @claude:

> ## Review fixes — `be6a544`
>
> Resolved 6 of the 8 findings plus the pagination tiebreaker. `pnpm verify` (666 hub + 104 frontend) and `pnpm lint` green.
>
> **Fixed**
> - **#1 paused-cast phantom play** — Sonos poll scrobble now gated on `s.state === "PLAYING"`.
> - **#2 sink-switch double-count** — `scrobbledRef` re-arms on track-id change only (one effect); a local↔Sonos switch keeps the same id and can't re-count.
> - **#3 near-end resume instant-scrobble** — threshold now measured against *this session's* progress (`absolutePos − sessionStart`), not absolute position.
> - **#4 lost play on backgrounded tab** — `handleEnded` scrobbles on completion as a fallback when `timeupdate` never crossed the threshold.
> - **#6 ignored `time` param** — scrobble now honors Subsonic `time` (epoch ms); `PlayEventService.record` gained `playedAtMs`. Offline/batched plays land at their real time.
> - **#7 stale comment** — rewritten to state the single recording path.
> - **Cut item (tiebreaker)** — `frequent`/`recent` `ORDER BY` now ends in `urg.id` for stable paging.
> - Plus the **duplicated threshold math** extracted to `scrobbleThresholdSec` in `lib/subsonic.ts`, shared by both playback paths (the root of #1/#3).
>
> New tests: paused-cast no-scrobble, near-end-resume no-scrobble, end-of-track fallback, threshold unit, `time` backfill.
>
> **Deferred as one coherent follow-up (#5 + #8)**
> Finding **#5** (folder-scoped Most Played ranks by cross-source counts) and **#8** (frequent/recent double-aggregates the whole history) are intertwined: fixing #5 alone — scoping the `playJoin` by instance while `annotatePlays`/`getAlbumStats` stays unscoped — would make the *ranking* and the *displayed* `playCount` disagree (exactly the split #8 flags). The right fix does both at once: select the `playJoin`'s `play_count`/`last_played` into the row, scope it via `play_events.source_instance_id`, and use those values for display instead of the second `getAlbumStats` query. I left it out of this commit to keep that as one reviewable change rather than a half-fix. Happy to do it next.
>
> Also noted, low severity: the AlbumsPage server-ordered set is enumerated in a few spots — worth a single `ViewSpec` flag when `type=recent` gets wired up.

benders:
FROM @claude:

> **#5 resolved — working as intended, no change.**
>
> Confirmed with the owner: play count is per *album* (release group), source-agnostic — the same album existing on multiple hubs is irrelevant, it's one album. The host-scoped "Most Played" views correctly use that per-album count and only filter *which* albums appear. So the unscoped `playJoin` is correct; no instance scoping wanted.
>
> This decouples **#8**, which is now a standalone performance cleanup (no behavior change): `getAlbumList2` frequent/recent still aggregates the play history twice per request — once in the `playJoin` for ranking (discarded) and again via `getAlbumStats` for display. Fix = select the `playJoin`'s `play_count`/`last_played` into the row and reuse them, dropping the second query.

benders:
FROM @claude:

> **#8 fixed — `21ea5f7`.** `getAlbumList2` frequent/recent now select the `playJoin`'s `play_count`/`last_played` into the row and reuse them for `playCount`/`played`, skipping the second `getAlbumStats` pass. History is scanned once per request, not twice. No behavior change — per-album, source-agnostic count is unchanged (#5 is working-as-intended). Tests extended to assert the reused values; `pnpm verify` (666 hub + 104 frontend) + `pnpm lint` green.
>
> That closes out the review: #1–#4, #6, #7, #8 + the pagination tiebreaker fixed; #5 confirmed no-change. Only the minor AlbumsPage cleanup (collapse the server-ordered enumeration into a `ViewSpec` flag) remains, best done when `type=recent` gets a view.

PR #250, opened 2026-07-12
feat(federation): peer lifecycle — disable, eviction tombstones, gossip propagation (v7)

### LINKED ISSUE #150: Tombstones & peer removal propagation
Follow-up to #147. Once gossip is in place, deletes don't propagate. Need:
- Soft-delete row (tombstone) in instances with signed delete attestation by inviter or peer-self
- Tombstones gossiped alongside peer entries; receivers honor them
- TTL/garbage-collection for old tombstones
- UI: admin removes a peer locally; option to issue signed removal that propagates

### LINKED ISSUE #151: Admin enable/disable for discovered peers

### LINKED ISSUE #157: fix(sync): instances.track_count inflated by duplicate upserts
FROM @claude:

> ## Problem
>
> `instances.track_count` for `id='local'` is inflated relative to the actual `instance_tracks` row count and the live Navidrome song count. Surfaces in the admin UI as a discrepancy between **"This Instance"** (2581) and **"Local Navidrome"** (2733), on a library where Navidrome reports 2582 present songs and `instance_tracks WHERE instance_id='local'` contains 2582 rows.
>
> ## Root cause
>
> `hub/src/library/sync-instance.ts:458` increments `result.trackCount++` on every `upsertTrack.run()` call. When Navidrome's album-detail iteration revisits the same song (multi-album appearances, compilations, multi-disc handling, overlapping album listings), the counter overshoots the unique track count.
>
> That inflated value is then written to `instances.track_count` at `hub/src/library/sync-instance.ts:476`:
>
> ```ts
> db.prepare(
>   "UPDATE instances SET ... track_count = ?, ... WHERE id = ?",
> ).run(syncMessage, result.trackCount, instanceId);
> ```
>
> `_deleteStale` correctly leaves only the distinct seen remote IDs in `instance_tracks`, so the data is fine — only the stored counter is wrong.
>
> ## Fix
>
> Use the de-duplicated count instead of the upsert-call counter. Either:
>
> - `seenTrackRemoteIds.size` (already maintained in scope), or
> - `SELECT COUNT(*) FROM instance_tracks WHERE instance_id = ?` after `_deleteStale`.
>
> Same likely applies to `result.albumCount` / `result.artistCount` — audit and fix in the same pass if affected.
>
> ## Verification
>
> - On a library with known duplicate-upsert conditions, `instances.track_count` should match `COUNT(*) FROM instance_tracks WHERE instance_id = 'local'` after sync.
> - Admin UI: "Local Navidrome" count should match Navidrome's `getScanStatus.count`.
> - Add/extend a sync-instance test that simulates the same song appearing under two albums and asserts `track_count` equals the distinct-track count.

benders:
Closes #150, closes #151. Refs #244 (Phases 1–3 of 5 — rate limiting #153 and backup/restore #152 are deferred to a follow-up PR per owner decision). Also fixes #157 in passing.

Adds the peer lifecycle state machine — `active` → `disabled` (reversible, local-only) → `tombstoned` (evicted, gossiped) — with local enforcement, an owner-only admin API + hub-admin UI, and cluster-wide eviction propagation as a **fully backwards-compatible** federation v7 wire change.

### Phase 1 — schema + local enforcement (0e2b297)

- `instances.lifecycle` (`active`|`disabled`|`tombstoned`) + `lifecycle_changed_at`, deliberately separate from the liveness `status` column health checks churn. Idempotent migration; `peer_tombstones` table for signed eviction records.
- Enforcement: `peer-auth.ts` refuses non-active peers with a uniform `403 {error:"forbidden"}` **after** signature verification (probes can't distinguish disabled ids from active ones — both see the usual 401s); `merge.ts` excludes non-active instances' rows (rows kept — re-enable needs no re-sync and restores the same deterministic unified ids, orphaned stars re-resolve); `sync.ts`/`auto-sync.ts` skip non-active peers before any sync op or health probe; `gossip.ts` refuses re-introduction of tombstoned ids and entries vouched by tombstoned inviters (disabled inviters stay trusted — local policy ≠ trust revocation, so the registry snapshot stays deliberately unfiltered for provenance pinning).
- **#157 drive-by fix**: `readNavidromeViaProxy` counted every `upsertTrack` visit, inflating `instances.track_count` when overlapping album listings (multi-disc/compilations) name the same song — now reports the deduplicated seen-id count, with a regression test.

### Phase 2 — admin API + UI (2771c76)

- Owner-only `POST /api/admin/hub/peers/:id/{disable,enable}` and `DELETE /api/admin/hub/peers/:id` (optional `{reason}`). All reload the `PeerRegistry` and run the merge pipeline inline so content disappears/reappears immediately. Tombstoned peers 409 on disable/enable (eviction isn't reversible that way); `local` 400s; unknown ids 404. DELETE is idempotent.
- `federation/tombstones.ts`: canonical Ed25519 payload `TOMBSTONE\n{instance_id}\n{removed_by}\n{created_at}` — `created_at` is generated in JS and stored verbatim so what's signed is exactly what's stored/gossiped. `createTombstone` + `verifyTombstone` both unit-tested (tamper cases included).
- hub-admin PeersSection: Disabled/Removed badges, Disable/Enable actions, confirm-guarded Remove.

### Phase 3 — tombstone gossip, federation v7 (this commit)

- **Wire**: `GET /federation/peers` gains a sibling `tombstones` array (rows relayed verbatim with the original remover's signature — never re-signed). **Additive and back-compat verified, not assumed**: v5/v6 parsers read only `body.peers` and validate known fields per entry, so the sibling field is invisible to them; the v7 client treats an absent field as an empty list. `FEDERATION_API_VERSION` 6→7, **floor stays 5**. Version-skew test covers both directions.
- **Ingest trust model** (documented in federation-api.md): tombstones verify against the remover's *pinned* pubkey; unknown removers are rejected, as are removers we've ourselves tombstoned (an evicted hub can't prune the mesh) and tombstones naming the receiving hub's own id. Any other admitted hub may evict any known instance — small mutually-trusted mesh, no quorum.
- **Re-admission**: an invitation whose `issued_at` postdates the tombstone's `created_at` clears it — enforced identically at the handshake path and the gossip path, and the tombstone is only cleared **after** the invitation signature verifies (a forged `issued_at` on an invalid entry can't un-evict anyone). Stale tombstones (older than the locally stored invitation) are rejected on ingest — this is what makes the mesh **converge** after a re-admission instead of flip-flopping the peer between evicted and re-admitted.
- Gossip ingest deliberately does not run the merge pipeline (it fires every sync cycle); content drops at the next merge, which the same sync runs immediately after gossip.

### Found during federation-suite verification (fixed here)

- **`/proxy/*` was not covered by the lifecycle gate.** `/proxy/*` has its own auth path (`proxy/auth.ts`), separate from `federation/peer-auth.ts` — the Phase 1 gate never applied to it, so an evicted peer could still stream and sync content. The suite's new eviction scenario caught it; the same post-signature uniform-403 gate now lives in both paths, with unit tests. Recorded in `docs/pitfalls.md`.
- **`proxy/auth.ts` version check compared against the *current* version, not the floor** — pre-existing, but the 6→7 bump would have cut v6 peers off from `/proxy/*` (exactly the compat break this PR must not make). Now compares against `MIN_FEDERATION_API_VERSION` (5), with floor tests.

### Verification

- `pnpm verify` green: hub **739/739** (61 files), frontend **115/115** (17 files); typecheck + `lint:boundary` clean; `pnpm lint` zero output.
- **Federation compat suite green** (Docker hub-a/b/c, compiled-JS path): py-sonic **84/84**, plus the new end-to-end eviction scenario — hub-a tombstones hub-c via the admin API → hub-b learns via gossip, tombstones it locally, drops its content from the merged catalog, and refuses hub-c's inbound requests. Scenario deliberately runs last (it degrades the mesh).
- Docs: `federation-api.md` v7 contract (wire shape, signing payload, verification rules, trust model, re-admission, changelog), `system-architecture.md` lifecycle section, `pitfalls.md` two new traps.

https://claude.ai/code/session_011Q3riZ78G26ohMgPDh8F7u


benders:
FROM @claude:
> Added end-to-end **re-admission** coverage to the federation suite (c516402): after the eviction scenario, hub-a re-invites hub-c with a fresh invitation, hub-b clears its tombstone via gossip on the next sync, hub-c's album returns to hub-b's merged catalog, and hub-c syncs successfully again.
>
> This surfaced one real gap on the invitee side of `POST /api/admin/hub/peers/accept`: the mirror INSERT of the inviter row wasn't idempotent (re-accepting from a known inviter 409'd), and there was no tombstone postdating check for a locally tombstoned *inviter*. Both fixed — the upsert flips `tombstoned → active` but deliberately leaves a locally `disabled` inviter disabled, and a stale invitation is refused 403 before any network call. Four new unit tests + a federation-api.md note.
>
> Full run green: `pnpm verify` (743 hub / 115 frontend), `pnpm lint` zero output, federation suite py-sonic 84/84 + eviction + re-admission acts.

PR #259, opened 2026-07-13
fix: restore local volume on unmute

### LINKED ISSUE #207: Clicking volume "speaker" icon when volume is at zero should set it low
Right now clicking the speaker sets it fairly high, nearly at max. Instead, it should restore to the previous position or set it at about halfway.

c436zhan:
## Summary
Restore the local player's previous non-zero volume when the speaker button unmutes from zero, instead of jumping to the hardcoded 0.8 value.

## Root cause
The local volume button used `setVolume(volume > 0 ? 0 : 0.8)`, so any unmute from zero ignored the user's previous slider position.

## Changes
- Track the last non-zero local volume while `PlayerBar` is mounted.
- Fall back to `0.5` when unmuting with no previous non-zero value.
- Add regression coverage for both previous-volume restore and no-history fallback.

## Testing
- `corepack pnpm --filter frontend test -- PlayerBar.test.tsx` (fails before the fix with 0.8, passes after: 22 tests / 123 frontend collected tests passed)
- `corepack pnpm --filter frontend typecheck` (pass)
- `corepack pnpm --filter frontend lint:boundary` (pass)
- `corepack pnpm --filter frontend lint` (pass)
- `corepack pnpm --filter frontend test` (pass: 18 files / 123 tests)
- `git diff --check` (pass)

Note: the root `pnpm verify` script cannot run directly in this environment because nested scripts call a non-executable `pnpm` shim, so I ran the equivalent commands through `corepack pnpm`. The unchanged hub unit suite was also checked; `test/admin-routes.test.ts` consistently times out in `admin — sync > POST /api/admin/hub/sync → returns 200 with local + peers result shape`, unrelated to this frontend-only volume change.

Fixes #207

### DIFF
--- a/frontend/src/components/player/PlayerBar.test.tsx
+++ b/frontend/src/components/player/PlayerBar.test.tsx
@@ -109,6 +109,52 @@ describe("PlayerBar render stability", () => {
   });
 });
 
+describe("PlayerBar local volume", () => {
+  it("restores the previous local volume when unmuting from zero", () => {
+    usePlayer.setState({
+      queue: [track("trk-1")],
+      currentIndex: 0,
+      volume: 0.36,
+    });
+
+    const { getByRole } = render(
+      <QueryClientProvider client={new QueryClient({ defaultOptions: { queries: { retry: false } } })}><MemoryRouter>
+        <PlayerBar />
+      </MemoryRouter></QueryClientProvider>,
+    );
+
+    act(() => {
+      fireEvent.click(getByRole("button", { name: "Mute" }));
+    });
+    expect(usePlayer.getState().volume).toBe(0);
+
+    act(() => {
+      fireEvent.click(getByRole("button", { name: "Unmute" }));
+    });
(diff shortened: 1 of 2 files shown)


benders:
FROM @claude:
> Thank you for this fix, @c436zhan — the root-cause diagnosis was spot on, and your component-level tests that drive the real Mute/Unmute buttons were the nicest test approach of the two competing PRs; we ported that style into the merged branch (4b5c706) with credit to this PR.
>
> Closing as a duplicate of #258, which was merged for issue #207. The deciding differences: the volume slider is square-law (`position = sqrt(volume)`), so the `0.5` raw fallback here renders at ~71% of the slider rather than halfway (the merged fix uses raw `0.25` = 50% position); #258 also extends remember/restore to the Sonos cast speaker button, and tracking every slider change in a ref meant a drag-to-zero left a near-silent last sample (e.g. `0.02`) as the restore level.

